PR #1, opened 2014-07-31
OPENJPA-2521 Cannot load entities from a different bundle in an OSGi environment

artaxerxe:
Fixed this issue, so that now you can load those entities even in a OSGi context.


mattrpav:
This issue appears dated, but the merge has no conflicts. If it makes sense, it'd be great to get it into 2.4.3

artaxerxe:
Hi all,

Is there any problem with my pull request? I still am using my version of OpenJPA in karaf because this isn't merged in the branch. Please let me know if I should do something.

Thanks,
Mura Andrei

rmannibucau:
Hi @artaxerxe , from what I see I think it still needs some work:

1. fix the formatting issue (imports)
2. ensure org.osgi.core scope is provided (optionally +optional too) since we don't want org.osgi.core in most environments
3. likely add a test ensuring it fixes the issue you mentionned (should be failing today and pass with the patch)
4. wonder if the bundle protocol handling shouldnt be added through a SPI and registered in PersistenceActivator only

wdyt?

artaxerxe:
Hi @rmannibucau ,

Of all points which you raised I'm not sure what you mean by 4, but I'll have a closer look and come back afterwards with questions if necessary.

Thanks.

rmannibucau:
@artaxerxe overall idea is to not do the OSGi handling in "main" code but keep it located in OSGi package if possible (not a huge requirement but would be saner overall for the project). One constraint is to be able to run without OSGi in all env, including security manager/javaagent ones and having an indirection makes it easier to guarantee but otherwise looks ok to me.

artaxerxe:
Hi @rmannibucau,

PR #2, opened 2015-07-20
OPENJPA-2601 Correct element name for query hint in orm.xml parser

kariem:
As described in [OPENJPA-2601](https://issues.apache.org/jira/browse/OPENJPA-2601), the `hint` element in `orm.xml` is not correctly parsed. This PR provides a test case and the relevant correction to `XMLPersistenceMetaDataParser`.


struberg:
Cosing, already applied a long time ago https://svn.apache.org/viewvc?view=revision&revision=1692088
txs for the patch again, Kariem!

PR #5, opened 2016-09-30
[OPENJPA-2669] Add karaf feature and adapt imports

cschneider:


openjpa-features/src/main/feature/feature.xml
```diff
@@ -0,0 +1,35 @@
...
+-->
+<features name="openjpa-${project.version}" xmlns="http://karaf.apache.org/xmlns/features/v1.3.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://karaf.apache.org/xmlns/features/v1.3.0 http://karaf.apache.org/xmlns/features/v1.3.0">
+
+    <feature name="openjpa" description="Apache OpenJPA 3 persistence engine support" version="${project.version}">
+        <details>Apache OpenJPA persistence engine</details>
+        <bundle dependency="true">mvn:org.apache.geronimo.specs/geronimo-jta_1.1_spec/1.1.1</bundle>
+        <bundle dependency="true">mvn:org.eclipse.persistence/javax.persistence/2.1.0</bundle>
+        <bundle dependency="true">mvn:org.apache.geronimo.specs/geronimo-annotation_1.0_spec/1.1.1</bundle>
+        <bundle dependency="true">mvn:org.apache.geronimo.specs/geronimo-el_1.0_spec/1.0.1</bundle>
+		<bundle dependency="true">mvn:commons-pool/commons-pool/1.6</bundle>
```
rmannibucau: we should likely move to dbcp2 today no?

rmannibucau:
@cschneider how hard would it be to have a unit test with pax to ensure it is not broken in the main build?

cschneider:
A pax exam test is on my todo list. I think it is crucial to make sure the feature still works in the future.

openjpa-features/src/main/feature/feature.xml
cschneider: Makes sense .. or not use require any pooling at all. I let pax-jdbc-config create a pooled DataSource

openjpa-features/src/main/feature/feature.xml
```diff
@@ -0,0 +1,35 @@
...
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+-->
+<features name="openjpa-${project.version}" xmlns="http://karaf.apache.org/xmlns/features/v1.3.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://karaf.apache.org/xmlns/features/v1.3.0 http://karaf.apache.org/xmlns/features/v1.3.0">
+
+    <feature name="openjpa" description="Apache OpenJPA 3 persistence engine support" version="${project.version}">
+        <details>Apache OpenJPA persistence engine</details>
+        <bundle dependency="true">mvn:org.apache.geronimo.specs/geronimo-jta_1.1_spec/1.1.1</bundle>
+        <bundle dependency="true">mvn:org.eclipse.persistence/javax.persistence/2.1.0</bundle>
```
cschneider: Well alpha-1 does not sound so trustworthy :-) I just checked that you created the bundle. Is it stable ? If yes it would be great if you could release a 1.0.0 version so people know it is safe to use.

openjpa-features/src/main/feature/feature.xml
rmannibucau: +1 to not handle it there

openjpa-features/src/main/feature/feature.xml
rmannibucau: alpha-1 for EE 7 spec was cause of this TCK thing only, I didn't check the OSGi part but that would be easy to enhance and stay consistent with karaf ecosystem I think.

gnodet:
Fwiw, validating the features with the Karaf 4 maven plugin ensure that the feature can be deployed correctly.  An integration test can certainly be done to actually test the bundles, but it's not necessary to verify that the bundles can be installed and resolved correctly.

rmannibucau:
Well I was more concerned about the runtime behavior than the installation (= ensure it deploys and then we can load properly classes, listeners etc... and that tccl is the right one).

Some work will get started soon with some lazy instantiation logic and I would like to avoid a half working state.

PR #7, opened 2016-11-19
OPENJPA-2554 JPA 2.1 - Schema Generation.

radcortez:


johnament:
I would drop the author tags.

radcortez:
My bad. I had them as a template for my IDE and forgot about it.

I'll remove them in a future patch.

PR #9, opened 2018-01-03
OPENJPA-2725 - fix ConcurrentModificationException during unrefed dependents cleanup

eran-levy:
In some situations there are state elements in _derefCache HashSet causing ConcurrentModificationException because set has been modified during iteration. 
Copy all elements into another set and calling the same deleteDeref(StateManagerImpl sm) function.

### DIFF
--- a/openjpa-kernel/src/main/java/org/apache/openjpa/kernel/BrokerImpl.java
+++ b/openjpa-kernel/src/main/java/org/apache/openjpa/kernel/BrokerImpl.java
@@ -2153,12 +2153,10 @@ public class BrokerImpl implements Broker, FindCallbacks, Cloneable, Serializabl
                 // mark for delete all elements in deref, otherwise in some situations it
                 // throws ConcurrentModificationException
                 statesMarkedForDelete.addAll(_derefCache);
-                if(statesMarkedForDelete.size()>0) {
-                    for (StateManagerImpl state: statesMarkedForDelete) {
-                        deleteDeref(state);
-                    }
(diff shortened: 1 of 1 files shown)


openjpa-kernel/src/main/java/org/apache/openjpa/kernel/BrokerImpl.java
```diff
@@ -2149,8 +2149,16 @@ protected void flush(int reason) {
             // dependents
             _flags |= FLAG_DEREFDELETING;
             if (flush && _derefCache != null && !_derefCache.isEmpty()) {
-                for (Iterator<StateManagerImpl> itr = _derefCache.iterator(); itr.hasNext();)
-                    deleteDeref(itr.next());
+                Set<StateManagerImpl> statesMarkedForDelete = new HashSet<>();
+                // mark for delete all elements in deref, otherwise in some situations it
+                // throws ConcurrentModificationException
+                statesMarkedForDelete.addAll(_derefCache);
+                if(statesMarkedForDelete.size()>0) {
```
ilgrosso: This if can be safely removed: the subsequent foreach will skip in case of empty set

ilgrosso:
Comment to link to OPENJPA-2725

openjpa-kernel/src/main/java/org/apache/openjpa/kernel/BrokerImpl.java
eran-levy: Correct, thanks. removed

eran-levy:
Fixed based on ilgrosso review

PR #17, opened 2018-08-17
[OPENJPA-2747] Upgrade to JPA 2.2 and use javax.persistence-api spec

jbonofre:


openjpa-integration/validation/pom.xml
```diff
@@ -95,11 +95,6 @@
                     <version>4.0.2.GA</version>
                     <scope>test</scope>
                     <exclusions>
-                        <!-- force usage of the geronimo-jpa_2.1_spec -->
```
rmannibucau: we should pby keep that excluded, no?

ilgrosso:
We have moved to geronimo's JPA specs 2.2 in 44aede26df24d0e0615309c7b31fc59a1479231e, closing this PR.
Thanks anyway @jbonofre !

PR #18, opened 2018-08-27
[OPENJPA-2567] various MySql text types support is added

solomax:


ilgrosso:
Thanks @solomax this PR lgtm and I would merge it both to `trunk` and `2_4_X`: @struberg @rmannibucau any objection?

rmannibucau:
+1

ilgrosso:
@solomax would you mind applying the same fixes to [MariaDBDictionary](https://github.com/apache/openjpa/blob/trunk/openjpa-jdbc/src/main/java/org/apache/openjpa/jdbc/sql/MariaDBDictionary.java)?

solomax:
Sure,
will do later tonight

solomax:
@ilgrosso MariaDb changes were added, previous PR had some files in wrong places (was done by `git apply svn.patch`), sorry for this

ilgrosso:
@solomax No troubles as we're on SVN, hence I could only grab the full diff for this PR and manually apply it to both `trunk` and `2.4.x` (with small modifications to restore Java 6 compatibility) branches.
Thanks!

ilgrosso:
We're also using GitBox at CXF, while at Syncope we're on the old Git / GitHub sync, still not finding enough reasons to move to the former, actually.

PR #19, opened 2018-08-30
[OPENJPA-2744] commons-pool is updated to the most recent version

solomax:


openjpa-all/pom.xml
```diff
@@ -31,7 +31,6 @@
         <version>3.0.1-SNAPSHOT</version>
     </parent>
 
-    <groupId>org.apache.openjpa</groupId>
```
ilgrosso: This change is not related to the topic, I don't see the reason to include it here, please revert.

openjpa-kernel/src/main/java/org/apache/openjpa/event/TCPRemoteCommitProvider.java
```diff
@@ -782,15 +816,17 @@ private HostAddress(String host)
...
-            _socketPool = new GenericObjectPool
-                (new SocketPoolableObjectFactory(), _maxActive,
-                    GenericObjectPool.WHEN_EXHAUSTED_BLOCK, -1);
+            _socketPool = new GenericObjectPool<>(new SocketPoolableObjectFactory(), cfg);
             _isAvailable = true;
         }
 
         private void setMaxActive(int maxActive) {
-            _socketPool.setMaxActive(maxActive);
+            _socketPool.setMaxTotal(maxActive);
```
ilgrosso: Why this change? Is `setMaxActive()`not available anymore?

openjpa-all/pom.xml
solomax: Maven raises the warning: `groupId` is duplicated
I'll move all these changes to separate PR

openjpa-kernel/src/main/java/org/apache/openjpa/event/TCPRemoteCommitProvider.java
solomax: Yep, it is replaced with `setMaxTotal`

ilgrosso: I see: would then make sense to rename the enclosing method accordingly?

solomax: I believe this is the decision OpenJPA dev's need to make ...
I tried to make changes minimal.

ilgrosso: Ok, I was comparing [1.6](https://commons.apache.org/proper/commons-pool/api-1.6/org/apache/commons/pool/impl/GenericObjectPool.html#setMaxActive(int))'s with [2.6.0](https://commons.apache.org/proper/commons-pool/api-2.6.0/index.html)'s javadocs and the description is effectively the same.

ilgrosso:
Thanks again @solomax !

Next targets could be:
* commons-collections -> commons-collections4
* commons-dbcp -> commons-dbcp2

WDYT? ;-)

solomax:
Already working on `commons-collections -> commons-collections4`
Found one serious issue so far, please check https://markmail.org/message/l72ev2vezesgbfj2
Maybe you can suggest something ...

PR #24, opened 2018-10-24
[OPENJPA-2754] maxActive -> maxTotal

solomax:
openjpa-jdbc module doesn't have any loggers
How I can raise warning in openjpa-jdbc/src/main/java/org/apache/openjpa/jdbc/schema/DBCPDriverDataSource.java ?

Would in be OK to add Logger? or are there better options?


openjpa-jdbc/src/main/java/org/apache/openjpa/jdbc/schema/DBCPDriverDataSource.java
```diff
@@ -194,8 +194,8 @@ private Properties updateDBCPProperties(Properties props) {
         if (hasKey(dbcpProps, "minIdle") == null) {
             dbcpProps.setProperty("minIdle", "0");
         }
-        if (hasKey(dbcpProps, "maxActive") == null) {
```
struberg: Probably we should keep both? In which case it should work with both old and new commons-dbcp libs, isn't?
Of course this needs some comment to explain the reason and the change which made it necessary.

openjpa-kernel/src/main/java/org/apache/openjpa/event/TCPRemoteCommitProvider.java
```diff
@@ -145,16 +145,16 @@ public int getRecoveryTimeMillis() {
      * The maximum number of sockets that this provider can
      * simetaneously open to each peer in the cluster.
      */
-    public void setMaxActive(int maxActive) {
```
struberg: I would _also_ keep the old method but log a warning which states that it might be removed and people shall use `setMaxTotal(int)` instead. Probably also deprecate it.

openjpa-jdbc/src/main/java/org/apache/openjpa/jdbc/schema/DBCPDriverDataSource.java
struberg: Hmmmm. Probably true. Those properties get applied via reflection anyway. So every method which exists on a ConnectionFactory will get invoked afair. But we should cross check this.

Probably we might even remove all that stuff? Because most of the pools have sensible defaults anyway, right?

rmannibucau: +1, it sounds worse to corrupt pool defaults and not enable the user to use the pool doc anymore

solomax: So I'll drop `SimpleDriverDataSource`, `DBCPDriverDataSource` and `AutoDriverDataSource`, and will check nothing is broken
Would it be sufficient?

solomax: Should `DriverDataSource` also be dropped?

solomax:
@struberg, @rmannibucau I have removed `DBCPDriverDataSource` and `AutoDriverDataSource` classes
Is it OK to merge?

rmannibucau:
@solomax hmm, overall looks ok except jdbc.DriverDataSource which lost dbcp option (so can't be a 3.0.1). Shouldnt we put the BasicDataSource class of dbcp here even if we don't deliver it? Idea of dropping the wrappers was to support it by reflection so theorically we can keep it, no?

solomax:
@rmannibucau I'm afraid I don't get it :(
`dbcp` should be one of the options for `jdbc.DriverDataSource`, 
but there should be no `DBCPDriverDataSource.java` and `AutoDriverDataSource.java`?
how it will work?

rmannibucau:
@solomax if you put o.a.commons.dbcp2...BasicDataSource it should work no?

solomax:
@rmannibucau got it :) will try to update PR :)

solomax:
Seems to be implemented
@rmannibucau could you please review latest commit?

rmannibucau:
Hmm, code still depends on dbcp which shouldnt be the case - was the goal of dropping datasource impl.

Overall it is ok to have dbcp as a string or setter mapping in a map of string/string (like "user" is mapped to "username" setter for this impl "xxxx") but the factory relies on the class so dbcp cant be optional still.

But yes, the "dbcp" config is what i had in mind

solomax:
@rmannibucau I see the only way to remove dpcp dependency: to revert my 2 last commits
So there will be `DBCPDriverDataSource` class

rmannibucau:
@solomax i'm not following that last statement. Let me try to explain a bit more my thought andlet see where we don't align:

1. in DataSourceFactory you can test rawDs.getClass().getName() (there are a few dbcp impl) or if you prefer to tolerate children we can alwaysuse  try { loadClass(base).isInstance(rawDs)} catch NoClassDefFoundError | ClassNotFoundException ignore) {} pattern
2. then we just have to maintain a mapping between properties and setters:

    private final Map<String, Map<String, String>> mappingPerClass = new HashMapString, Map<String, String>() {{
         put("dbcp class name", new HashMap<String, String>() {{
            put("connectionUsername", "username");
            put("driver", "driverClassName");
         }}
    }}


etc... (feel free to directly use setter names instead of field name if more relevant)

this way we are loosely coupled and don't require at all dbcp. The alternative is to make dbcp optional and have a small indirection "DataSourceFactoryHandler" with accept(Object instance) and map(properties)

does it sounds better this way?

rmannibucau:
@solomax works for me

side note for another PR - but dont want to forget it: why do we need this instead of creating a set of properties (a map string -> object) and allowing the user to configure it with our dsl "dbcp(username=user,password=connectionpassword)" etc

solomax:
Could you please mark JIRA as resolved?

PR #40, opened 2019-04-24
[OPENJPA-2668] fix for execution of a CriteriaQuery leaking into CriteriaQuery state

michaelwiles:
for [OPENJPA-2668](https://issues.apache.org/jira/browse/OPENJPA-2668)
added copy method to Expressons.In and use that to return the expression
so that the original is not changed.

Not sure if there's a better way to do this. This is quite how can I say, simple?

This is the only Expression that I could find that dirties the original CriteriaQuery and I'm fairly convinced that this shouldn't happen. 

Prior to this fix, executing the test would change the query and replace the parameter with the hydrated parameter values.

### DIFF
--- a/openjpa-all/pom.xml
+++ b/openjpa-all/pom.xml
@@ -28,7 +28,7 @@
     <parent>
         <groupId>org.apache.openjpa</groupId>
         <artifactId>openjpa-parent</artifactId>
-        <version>3.1.0</version>
+        <version>3.1.1-SNAPSHOT</version>
     </parent>
 
     <artifactId>openjpa-all</artifactId>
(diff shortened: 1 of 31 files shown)


ilgrosso:
Thanks @michaelwiles for your contribution.
Would you please adjust `openjpa-persistence/src/main/java/org/apache/openjpa/persistence/criteria/Expressions.java` so that only effective modifications are submitted? I see large changes due to formatting, ATM.

michaelwiles:
hmmm... I'm going to create another pull request... this one is too broken to fix.

michaelwiles:
remade [here](https://github.com/apache/openjpa/pull/41)

PR #45, opened 2019-05-14
[master] OPENJPA-2767: Incomplete ValueMapDiscriminatorStrategy cache and Meta…

dazey3:
https://issues.apache.org/jira/browse/OPENJPA-2767

Signed-off-by: Will Dazey <dazeydev.3@gmail.com>

### DIFF
--- a/openjpa-kernel/src/main/java/org/apache/openjpa/meta/MetaDataRepository.java
+++ b/openjpa-kernel/src/main/java/org/apache/openjpa/meta/MetaDataRepository.java
@@ -126,7 +126,8 @@ public class MetaDataRepository implements PCRegistry.RegisterClassListener, Con
     private Map<Class<?>, Class<?>> _metamodel = Collections.synchronizedMap(new HashMap<Class<?>, Class<?>>());
 
     // map of classes to lists of their subclasses
-    private Map<Class<?>, List<Class<?>>> _subs = Collections.synchronizedMap(new HashMap<Class<?>, List<Class<?>>>());
+    private Map<Class<?>, Collection<Class<?>>> _subs =
+            Collections.synchronizedMap(new HashMap<Class<?>, Collection<Class<?>>>());
 
     // xml mapping
     protected final XMLMetaData[] EMPTY_XMLMETAS;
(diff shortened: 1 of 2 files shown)


openjpa-kernel/src/main/java/org/apache/openjpa/meta/MetaDataRepository.java
```diff
@@ -1625,19 +1626,20 @@ private void loadRegisteredClassMetaData(ClassLoader envLoader) {
...
     /**
-     * Updates our datastructures with the latest registered classes.
+     * Updates our data structures with the latest registered classes.
+     * 
+     * This method is synchronized to make sure that all data structures are fully updated
+     *  before other threads attempt to call this method
      */
-    Class<?>[] processRegisteredClasses(ClassLoader envLoader) {
-        if (_registered.isEmpty())
+    synchronized Class<?>[] processRegisteredClasses(ClassLoader envLoader) {
```
ilgrosso: Any chance to avoid synchronization at method level?

dazey3: I had written a possible implementation using read/write locks similar to my ValueMapDiscriminator impl, but so much of the code ended up being write locked that it almost seemed pointless to have that fine grained locking. Almost all the time, `_registered` is empty anyway, so little to no blocking will occur.
The issue really ends up being that all of processRegisteredClasses needs to block until it's done processing all of the classes out of the temporary array. I realize the temp array was added to reduce blocking on the `_registered` array in the first place, but that only opens a window for other threads to start initializing other caches WHILE the MetaDataRepository is still processing and building. If the ORM is sufficiently large, I was seeing this processing taking up to 700ms. During that time, reads are potentially dirty from `_subs`.
That being said, if there is a strong performance concern from synchonizing the method (something that may not be as troubling nowadays with newer jdks), I can re-investigate implementing locks. This change was just infinitely cleaner.

ilgrosso: @dazey3 I see: given what you say, method synchronization could be acceptable then - I was just wondering why...

dazey3: @ilgrosso I understand the concern and I appreciate all the review comments I can get! Thank you for taking a look.

PR #49, opened 2019-05-20
[2.2.x] Convert build from SVN to GIT

dazey3:
I am updating the 2.2.x so that we use the git version as the revision value instead of subversion. I decided to go with the git short hash instead of the full hash. I can port this to master if it's wanted, but I wasnt planning on it.

I ended up removing the SVNUtil tests that we used to have because there doesnt seem to be a need anymore for String manipulation and having a util that manages it. Now we just get the git hash.

Signed-off-by: Will Dazey <dazeydev.3@gmail.com>


ilgrosso:
@dazey3 I would merge almost the same changes - except for the root `pom.xml` where things should be already as proposed - onto `master` branch as well.

ilgrosso:
@dazey3 FYI I cherry-picked this commit to branch `2.2.x` as 711ad2145ac1efdf7ae0edcf2c86f7bc03313b56

PR #51, opened 2019-06-20
[2.2.x] OPENJPA-2790: Convert build from SVN to GIT

jgrassel:
https://issues.apache.org/jira/browse/OPENJPA-2790

This change ensures that git hashes have a length of 7 (observed different default lengths between different versions of git), and restores the "M" at the end of the revision number to signify "dirty" OpenJPA builds.

### DIFF
--- a/openjpa-lib/pom.xml
+++ b/openjpa-lib/pom.xml
@@ -96,17 +96,34 @@
                         <phase>compile</phase>
                         <configuration>
                             <target>
+                                <!-- SVN would have appended 'M' to the end of the revision number if there were
+                                     uncommitted changes.  Git doesn't do that, so we have to work a bit to
+                                     emulate that behavior.  -->
+                                <echo>Determining if there are uncommitted changes...</echo>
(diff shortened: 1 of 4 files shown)


openjpa-lib/pom.xml
```diff
@@ -96,17 +96,29 @@
...
                         <configuration>
                             <target>
+                                <!-- SVN would have appended 'M' to the end of the revision number if there were
+                                     uncommitted changes.  Git doesn't do that, so we have to work a bit to
+                                     emulate that behavior.  -->
+                                <echo>Determining if there are uncommitted changes...</echo>
+                                <exec outputproperty="git.diff.exitval" resultproperty="diff.state" failonerror="false" failifexecutionfails="false" executable="git">
+                                    <arg line="diff --quiet HEAD" />
+                                </exec>
+                                <condition property="uncommitted.changes" value="M">
```
dazey3: How about?: 
```java
<condition property="uncommitted.changes" value="M" else="">
```

jgrassel: That works for me too.

jgrassel: Done

PR #59, opened 2020-04-05
[OPENJPA-2808] indicies are dropped in schema tool

solomax:


rmannibucau:
Hi @solomax ,

1. is it possible to harness that with a test?
2. wonder if it needs a flag since it can change old behavior in test suites, wdyt?

otherwise looks like an old due debt ;), thanks a lot.

solomax:
Hello @rmannibucau, thanks for review
1. Can this `TestSchemaGenerator` test be considered as good example?
2. There is `SchemaTool.Flags.indexes` already, is it not enough?

Sorry for the long response

PR #62, opened 2020-06-09
Checking array size before access to avoid ArrayIndexOutOfBoundsException

ilgrosso:
I recently found that, under some circumstances that I am unfortunately not able to reproduce, I started getting `ArrayIndexOutOfBoundsException`from the `PrimaryRow` class.

The attached fix is simply checking array size before attempting to access.

### DIFF
--- a/openjpa-jdbc/src/main/java/org/apache/openjpa/jdbc/sql/PrimaryRow.java
+++ b/openjpa-jdbc/src/main/java/org/apache/openjpa/jdbc/sql/PrimaryRow.java
@@ -147,7 +147,7 @@ public class PrimaryRow
      * Return the I/O information for the given set foreign key.
      */
     public ColumnIO getForeignKeyIO(ForeignKey fk) {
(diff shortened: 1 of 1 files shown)


ilgrosso:
Thanks @solomax !

rmannibucau:
Is there a jira about it? (for the changelog)
When can it arrive - = any way to reproduce it in a test? (when DB already exists and JPA model is not 1-1?)

ilgrosso:
> Is there a jira about it? (for the changelog)

No, I would have added it to the commit message, in case; I thought it was not needed, being just an additional check to avoid an exception

> When can it arrive - = any way to reproduce it in a test?

As I said above, I am not able to reproduce it systematically but, looking at the code, I thought it was safer any way to add a check, so I did.

rmannibucau:
Hmm, concretely it shouldn't be possible and it is safe to directly access it since we computed the model before so to me - and I don't fully get the source yet - it looks like it hides the real issue more than fixing it, no?

PR #63, opened 2020-07-13
OPENJPA-2816 Add HerdDB DBDictionary

eolivelli:
- implement HerdDBDictionary
- add autodiscovery of HerdDB


ilgrosso:
...no tests at all?

rmannibucau:
@ilgrosso IT are coming but are IT (for ext db they are only meaningful if covering 100% of db version otherwise it is like having no test for all other versions ;)). We also have issues with our IT as mentionned by Mark on the list and it hits here so need a few days.

ilgrosso:
We do have treats for other in-memory dbms, I was thinking to include the news one.
...and "some test" is always better than "no tests".

rmannibucau:
@ilgrosso agree, point was more that IMHO it is more important to make it its way on master than to get an IT which is not straight forward yet (a mix between both the DB and our own lacks of our suite). To be concrete I expect a snapshot to be grabbable in the coming days for CI to catch/test up with the newt config so test are pushed to another PR. Hope it makes sense.

PR #66, opened 2020-07-15
Fix SelectImpl#getColumnAlias in case of DBDictionary that use delimiters aggressively

eolivelli:
This is a draft PR.
I am trying to run all of the integration tests against HerdDB.
HerdDBDictionary enables aggressive use of delimiters, in order to not have problems with the the long list of reserved words in Apache Calcite.

TestPersistence fails and I am suggesting this fix.
If the fix is correct, can you please suggest me where to add a test case ?

Please note that this is not the only fix needed to run integration tests with HerdDB, I will follow up with other PRs or requests


rmannibucau:
This is what does the getColumnAlias in the enclosing class so it looks good to me, just requires to check all IT on all DB but I'm quite confident.

eolivelli:
@rmannibucau  @struberg  I am closing this PR, superseded by #67

eolivelli:
GitHub does not allow me to reopen this PR, created #68 68

PR #68, opened 2020-07-16
Fix SelectImpl#getColumnAlias in case of DBDictionary that use delimiters aggressively

eolivelli:
This is a draft PR.
I am trying to run all of the integration tests against HerdDB.
HerdDBDictionary enables aggressive use of delimiters, in order to not have problems with the the long list of reserved words in Apache Calcite.

TestPersistence fails and I am suggesting this fix.
If the fix is correct, can you please suggest me where to add a test case ?

I am reopening #66, already reviewed by @struberg


eolivelli:
sorry for the noise, this patch makes sense only with #69 
so I have merged the two patches

PR #69, opened 2020-07-16
OPENJPA-2818 DBDictionary > delimitIdentifiers = true does not work

eolivelli:
While working at HerdDBDictionary I figured out that integration tests do not work because of two problems:
- a bug in SelectImpl#getColumnAlias, in which we are not applying delimiters to the column name
-ensure that Table#getColumn is not case sensitive and not sensitive to delimiters (store the key in UPPER case and without delimiters)
- add new test case TestDelimitIdentifiers.java that forces the use of delimiters and covers the changes
- fix herddb detection and add test case about DBDictionaryFactory

### DIFF
--- a/openjpa-jdbc/src/main/java/org/apache/openjpa/jdbc/schema/SchemaTool.java
+++ b/openjpa-jdbc/src/main/java/org/apache/openjpa/jdbc/schema/SchemaTool.java
@@ -604,6 +604,9 @@ public class SchemaTool {
                 for (int k = 0; k < cols.length; k++) {
                     if (dbTable != null) {
                         DBIdentifier colName = cols[k].getIdentifier();
+                        if (_dict.getDelimitIdentifiers()) {
+                            colName = DBIdentifier.removeDelimiters(colName);
+                        }
                         col = dbTable.getColumn(colName);
                         if (col == null) {
                             if (addColumn(cols[k]))
--- a/openjpa-jdbc/src/main/java/org/apache/openjpa/jdbc/schema/SchemaTool.java
+++ b/openjpa-jdbc/src/main/java/org/apache/openjpa/jdbc/schema/SchemaTool.java
@@ -956,7 +956,7 @@ public class SchemaTool {
                     if (!dbCols[k].getIdentifier().getName().equals(_dict.getIdentityColumnName()) &&
                         !tabs[j].containsColumn(dbCols[k]))
                         continue tables;
-                    }
+                }
                 drops.add(tabs[j]);
             }
         }
--- a/openjpa-jdbc/src/main/java/org/apache/openjpa/jdbc/schema/SchemaTool.java
+++ b/openjpa-jdbc/src/main/java/org/apache/openjpa/jdbc/schema/SchemaTool.java
@@ -40,6 +40,8 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.Set;
 
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 import javax.sql.DataSource;
 
 import org.apache.openjpa.conf.OpenJPAConfiguration;
@@ -604,9 +606,6 @@ public class SchemaTool {
                 for (int k = 0; k < cols.length; k++) {
                     if (dbTable != null) {
                         DBIdentifier colName = cols[k].getIdentifier();
-                        if (_dict.getDelimitIdentifiers()) {
-                            colName = DBIdentifier.removeDelimiters(colName);
-                        }
                         col = dbTable.getColumn(colName);
                         if (col == null) {
                             if (addColumn(cols[k]))
@@ -957,7 +956,7 @@ public class SchemaTool {
                     if (!dbCols[k].getIdentifier().getName().equals(_dict.getIdentityColumnName()) &&
                         !tabs[j].containsColumn(dbCols[k]))
                         continue tables;
-                }
+                    }
                 drops.add(tabs[j]);
             }
         }
@@ -1033,10 +1032,7 @@ public class SchemaTool {
                             continue;
 
                         if (dropColumn(cols[k])) {
-                            if (dbTable != null)
-                                dbTable.removeColumn(col);
-                            else
-                                _log.warn(_loc.get("drop-col", cols[k], tabs[j]));
+                           dbTable.removeColumn(col);
                         }
                     }
                 }
--- a/openjpa-jdbc/src/main/java/org/apache/openjpa/jdbc/schema/Table.java
+++ b/openjpa-jdbc/src/main/java/org/apache/openjpa/jdbc/schema/Table.java
@@ -38,6 +38,7 @@ import org.apache.openjpa.lib.meta.SourceTracker;
  * @author Abe White
  * @author Stephen Kim
  */
+
 public class Table
     extends NameSet
     implements Comparable<Object>, SourceTracker {
@@ -321,12 +322,12 @@ public class Table
         if (_colMap == null) {
             return new String[0];
         }
-        DBIdentifier[] sNames = _colMap
+        return _colMap
                 .values()
                 .stream()
                 .map(Column::getIdentifier)
-                .toArray(DBIdentifier[]::new);
-        return DBIdentifier.toStringArray(sNames);
+                .map(DBIdentifier::getName)
+                .toArray(String[]::new);
     }
 
     /**
--- a/openjpa-jdbc/src/main/java/org/apache/openjpa/jdbc/schema/Table.java
+++ b/openjpa-jdbc/src/main/java/org/apache/openjpa/jdbc/schema/Table.java
@@ -24,8 +24,10 @@ import java.util.Collection;
 import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.Map;
+import java.util.Set;
 import java.util.TreeMap;
 
+import java.util.stream.Collectors;
 import org.apache.openjpa.jdbc.identifier.DBIdentifier;
 import org.apache.openjpa.jdbc.identifier.DBIdentifier.DBIdentifierType;
 import org.apache.openjpa.jdbc.identifier.QualifiedDBIdentifier;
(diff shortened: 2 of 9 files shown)


eolivelli:
I will be happy to add tests in case the direction of this patch is good

rmannibucau:
@eolivelli looks "expected" to me but it needs harnessing for both changes too IMHO

eolivelli:
@rmannibucau the patch became bigger as soon as I tried to add test cases.
Still it is not complete, I have to add a test case of the  SelectImpl#getColumnAlias case

This patch also is fixing HerdDBDictionary, if you prefer I can separate that change in HerdDBDictionary ticket

rmannibucau:
@eolivelli it is ok to fix herddb dict there since it is fully related IMHO. The first part of the diff is suspicious (schematool diff seems unnecessary and getColumnNames change is unlikely too). Would also be neat to not drop delimiters if the dictionnary does not use them (normalizeColumnKey). If it helps I think we can add a flag in DBIdentifier to notify it is delimited or not and have both the delimited and not delimited values (raw name vs statement name).

openjpa-jdbc/src/main/java/org/apache/openjpa/jdbc/schema/SchemaTool.java
```diff
@@ -1030,10 +1032,7 @@ private void drop(SchemaGroup db, SchemaGroup repos, boolean considerDatabaseSta
                             continue;
 
                         if (dropColumn(cols[k])) {
-                            if (dbTable != null)
```
eolivelli, line 1033: dbTable is always not null here, we have a null check 4 lines above,
I can revert if you want

openjpa-jdbc/src/main/java/org/apache/openjpa/jdbc/schema/Table.java
```diff
@@ -312,11 +315,20 @@ public String getResourceName() {
...
 
+    /**
+     * Return the list of column names, used only for informative (error) messages.
+     * @return
+     */
     public String[] getColumnNames() {
         if (_colMap == null) {
             return new String[0];
         }
-        DBIdentifier[] sNames = _colMap.keySet().toArray(new DBIdentifier[_colMap.size()]);
```
eolivelli, line 319: `_colMap.keySet()` contains an internal representation of column names (all uppercased)
this method is used only for error messages and IMHO it is better to have a consistent handling of this internal data structure

openjpa-jdbc/src/main/java/org/apache/openjpa/jdbc/schema/Table.java
```diff
@@ -329,10 +341,19 @@ public Column getColumn(String name) {
...
+        DBIdentifier key = normalizeColumnKey(name);
+        return _colMap.get(key);
+    }
+
+    public Column getColumn(DBIdentifier name) {
+        return internalGetColumn(name);
+    }
+
+    private DBIdentifier normalizeColumnKey(DBIdentifier name) {
+        return DBIdentifier.removeDelimiters(DBIdentifier.toUpper(name, true));
```
eolivelli, line 356: @rmannibucau about applying removeDelimiters only when needed:
my original fix was to handle this only in SchemaTool, but just adding the new test case showed that  the situation is more complex.
Table class does not have a reference to the DBDictionary and we can't change the signature of all methods, it is kind of Public API that can be overridden by custom DBDictionary implementations

Tests are also failing if you switch to `DBIdentifier.toUpper(DBIdentifier.removeDelimiters(name), true);`

IMHO we could also probably clean up some of those methods in order to reduce useless allocation, but it can be a separate follow up patch.

openjpa-jdbc/src/main/java/org/apache/openjpa/jdbc/schema/Table.java
rmannibucau: _colMap
                .values()
                .stream()
                .map(Column::getIdentifier)
                .toArray(DBIdentifier[]::new);

openjpa-jdbc/src/main/java/org/apache/openjpa/jdbc/schema/Table.java
rmannibucau: the identifier should always (i know a few cases where it is not the case but shouldnt impact you or means it must be fixed) be created from the dict so just inject that state in the identifier? Makes sense to me to know if the identifier had been quoted or not and does not require a single signature change in methods.

openjpa-jdbc/src/main/java/org/apache/openjpa/jdbc/schema/Table.java
eolivelli: @rmannibucau 
I can't get your suggestion
If I change this to:
```
_colMap
.values()
.stream()
.map(Column::getIdentifier)
.toArray(DBIdentifier[]::new);
```
I get a DBIdentifier[] and not a String[]

openjpa-jdbc/src/main/java/org/apache/openjpa/jdbc/schema/Table.java
eolivelli: > Makes sense to me to know if the identifier had been quoted or not and does not require a single signature change in methods
I would like this change but it is not trivial, as I see that sometimes we are manipulating the internal `_name` of DBIdentifier, and tracking when it is already delimited or not seems to me an hard task.
It can be done in a follow up patch. I will be happy to follow up in this direction

openjpa-jdbc/src/main/java/org/apache/openjpa/jdbc/schema/Table.java
eolivelli: done

openjpa-jdbc/src/main/java/org/apache/openjpa/jdbc/sql/SelectImpl.java
```diff
@@ -2711,7 +2711,7 @@ private String getColumnAlias(Column col, PathJoins pj) {
                 return alias + "_" + col;
             }
             alias = SelectImpl.toAlias(_sel.getTableIndex(col.getTable(), pj, false));
-            return (alias == null) ? null : alias + "." + col;
+            return (alias == null) ? null : alias + "." + _sel._dict.getNamingUtil().toDBName(col.toString());
```
eolivelli, line 2714: the new test exercise this line.
I must be honest, with Derby it does not fail the test,  even without this line, but without this change the integration tests with HerdDB fail

openjpa-jdbc/src/main/java/org/apache/openjpa/jdbc/schema/Table.java
rmannibucau: Proposal was just to not to toList().toArray which allocates twice for no real reason the same structure. Probably missed a .map in my snippet but the double allocation should be avoided IMHO.

eolivelli: yes I understood and I have updated the code, thanks

openjpa-jdbc/src/main/java/org/apache/openjpa/jdbc/schema/Table.java
rmannibucau: It is not really about the overhead (should be a "contains" in terms of complexity for most cases) but more about being fully deterministic and not "random".

Why is it hard?
Tables can be created in:
- openjpa-jdbc/src/main/java/org/apache/openjpa/jdbc/sql/DBDictionary.java#newTable
- openjpa-jdbc/src/main/java/org/apache/openjpa/jdbc/sql/DBDictionary.java#getValidColumnName
- org.apache.openjpa.jdbc.schema.Schema#addTable
- org.apache.openjpa.jdbc.schema.SchemaGroup#newTable

Now if you take 1 more level you always have the dictionnary handy so if simpler we can just add the dict in the table.

Alternative I had in mind was just to handle it in org.apache.openjpa.jdbc.sql.DBDictionary#getValidColumnName(org.apache.openjpa.jdbc.identifier.DBIdentifier, org.apache.openjpa.jdbc.schema.Table, boolean) in herddb dict and set the flag there.
It is not about tracking it is delimited or not but tracking that the dict is in delimited mode or not, then you would do 
 if identifier.delimited then removeDelimited() else noop.

openjpa-jdbc/src/main/java/org/apache/openjpa/jdbc/sql/SelectImpl.java
rmannibucau: you can likely create an assert on the select string directly - even using internals to instantiate it if it helps.

openjpa-jdbc/src/main/java/org/apache/openjpa/jdbc/schema/Table.java
rmannibucau: Hmm, maybe diff is not up to date but you still allocate 2 arrays, first one is a DBIdentifier array instead of an ArrayList but it is ~the same size (I'm thinking to the case we have > 200 columns in the db and it happens) so I would just do:

_colMap
.values()
.stream()
.map(Column::getIdentifier)
.map(DBIdentifier::getName)
.toArray(String[]::new);

eolivelli:
all tests passed locally

eolivelli:
now I see why you asked for 'install'
the ant plugin needs the jars into the maven repository

openjpa-jdbc/src/main/java/org/apache/openjpa/jdbc/schema/Table.java
rmannibucau: Still a double alloc no?

rmannibucau:
Except the double alloc it looks good to me, @struberg want to have a quick look before we merge?

eolivelli:
all tests are passing locally on this branch

rmannibucau:
Once double allocation (https://github.com/apache/openjpa/pull/69/files#diff-59344a422210f668cbeb574dc22e8339R324) is fixed it can be merged for me

eolivelli:
@rmannibucau  let's see now github validation in action

rmannibucau:
Mergeable for me (i cant do it right now so if any dev with the perms agree and comes feel free to hit the green button)

ilgrosso:
whoever and does the merge, please do "squash & merge" :-)

eolivelli:
@ilgrosso or @rmannibucau can you please merge this patch.
I am going to work on the follow up work about storing into DBIdentifier a property that tells if the id has been already delimited

ilgrosso:
@eolivelli done!
Thanks for your contribution.

eolivelli:
@ilgrosso thanks to you guys for providing OpenJPA to the public
starting on https://issues.apache.org/jira/browse/OPENJPA-2820 soon

PR #70, opened 2020-07-17
OPENJPA-2819 Add simple GitHub Actions validation for Pull Requests

eolivelli:
Add simple pull request validation.
Starting from the github actions template https://github.com/actions/starter-workflows/blob/master/ci/maven.yml

### DIFF
--- a/.github/workflows/pr-validation.yml
+++ b/.github/workflows/pr-validation.yml
@@ -33,6 +33,6 @@ jobs:
           key: 'cache'
           restore-keys: 'cache'
       - name: 'Build with Maven'
-        run: mvn -B verify --file pom.xml
+        run: mvn -B package --file pom.xml
       - name: 'Remove Snapshots Before Caching'
         run: find ~/.m2 -name '*SNAPSHOT' | xargs rm -Rf
--- a/.github/workflows/pr-validation.yml
+++ b/.github/workflows/pr-validation.yml
@@ -1,0 +1,42 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more contributor
+# license agreements.  See the NOTICE file distributed with this work for additional
+# information regarding copyright ownership.  The ASF licenses this file to you
+# under the Apache License, Version 2.0 (the # "License"); you may not use this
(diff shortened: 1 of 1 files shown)


rmannibucau:
Looks good to me now, only thing to decide is if we keep on push or not (does not hurt I think and can be good for us).
Waiting for a passing build then I think we can merge.


cc @ilgrosso @struberg want to have a look?

ilgrosso:
@eolivelli why "only" `package`? Why not `install`?

@rmannibucau +1 to add it to push as well.

eolivelli:
@ilgrosso thanks for your review

There is not need to 'install'.
With 'install' Maven copies the jar to .m2.

It looks like the build works without it.
Probably 'verify' could work ad well

ilgrosso:
Agree then, go on with verify

eolivelli:
@ilgrosso done

eolivelli:
If you commit this change github actions will automatically be activated probably

eolivelli:
The build fails with verify

[ERROR] /home/runner/work/openjpa/openjpa/openjpa-examples/openbooks/build.xml:117: *** Error:
[ERROR] 		   The directory for OpenJPA libraries can not be located at /home/runner/.m2/repository/org/apache/openjpa/openjpa-all/3.1.3-SNAPSHOT.
[ERROR] 		   Make sure openjpa.lib property value is correct in build.properties file.
[ERROR] around Ant part ...<ant

eolivelli:
this is a demo run:
https://github.com/eolivelli/openjpa/runs/883352848

eolivelli:
It looks like the ant plugin looks for openjpa lib into local Maven repository.
I will change the goal to 'install' as @ilgrosso suggested

eolivelli:
try again:
https://github.com/eolivelli/openjpa/runs/883449807

eolivelli:
@ilgrosso @rmannibucau @struberg an you please commit this patch ?
it looks like it works very well
https://github.com/eolivelli/openjpa/runs/883449807

rmannibucau:
+1 from me

PR #75, opened 2020-09-28
OPENJPA-2832 DROP COLUMN does not use delimiters and always add double quotes

eolivelli:


rmannibucau:
think we should do that for all dictionary methods before merging to avoid a PR ping pong, wdyt?

rmannibucau:
Merging but would be great to review default dictionary more globally since some other columns need it too

PR #76, opened 2020-10-27
OPENJPA-2836 HerdDBDictionary: enable ForeignKeys and Unique Indexes

eolivelli:
Drop these lines from HerdDBDictionary, as HerdDB 0.21.0+ supports Unique Indexes and HerdDB 0.22.0+ supports ForeignKeys with "ON DELETE CASCADE" action.

```
        supportsForeignKeys = false;
        supportsUniqueConstraints = false;
        supportsCascadeDeleteAction = false;
```

### DIFF
--- a/openjpa-jdbc/src/main/java/org/apache/openjpa/jdbc/sql/HerdDBDictionary.java
+++ b/openjpa-jdbc/src/main/java/org/apache/openjpa/jdbc/sql/HerdDBDictionary.java
@@ -33,6 +33,7 @@ public class HerdDBDictionary
         delimitedCase = SCHEMA_CASE_LOWER;
 
         supportsCascadeUpdateAction = false;
+        supportsDeferredConstraints = false;
 
         // make OpenJPA escape everything, because Apache Calcite has a lot of reserved words, like 'User', 'Value'...
         setDelimitIdentifiers(true);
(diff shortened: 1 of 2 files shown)


eolivelli:
@rmannibucau the build failed, but the error is not related to my patch

```
Error:  Failures: 
Error:    TestPropertiesMethods>AbstractPersistenceTestCase.run:206->AbstractPersistenceTestCase.runBare:541->AbstractPersistenceTestCase.runBare:565->AbstractPersistenceTestCase.runTest:578->testChangePropertyValue:236->assertProperty:277 openjpa.IgnoreChanges value does not match expected:<true> but was:<false>
```

rmannibucau:
@eolivelli I will take care of this one but you should add tests for the 3 changes you do too ;)

eolivelli:
@rmannibucau I am not sure about the type of tests to add here.
Shall I add a unit test that instantiates an HerdDBDictionary and checks the default values of those fields ?

rmannibucau:
@eolivelli basically you should check that when these features are used they lead to issuing the right SQL statement for the related columns. You can either use the dict directly (likely easier) or generate some DDL using SchemaTool or so using HerdDB dict and check it is the expected ones.

eolivelli:
Thanks foe your hint.
I will add new tests

eolivelli:
@rmannibucau PR updated

rmannibucau:
LGTM, @struberg or @ilgrosso do you want to have a look too before we merge?

ilgrosso:
LGTM

PR #78, opened 2021-04-06
change newline encoding to silent checkstyle

peutch:
Hi,
I just have done this:
`find . -name "*.java" -exec file {} \; | grep CRLF | cut -d ':' -f 1 | while read f ; do dos2unix "$f" ; done`

Moreover trying to integrate openjpa source in my dev env (Eclipse) I have other troubles like
for the following class:
`./openjpa-lib/src/main/java/org/apache/openjpa/lib/conf/ClassListValue.java`
that requires Java 9 at least for source code, isn't it?


peutch:
Regarding ClassListValue it is in fact another problem.

solomax:
Hello @peutch,

I've just run `mvn clean install` at `Ubuntu 20.04`

```
Apache Maven 3.8.1 (05c21c65bdfed0f71a2f2ada8b84da59348c4c5d)
Maven home: /opt/maven
Java version: 1.8.0_282, vendor: Private Build, runtime: /usr/lib/jvm/java-8-openjdk-amd64/jre
Default locale: en_US, platform encoding: UTF-8
OS name: "linux", version: "5.4.0-70-generic", arch: "amd64", family: "unix"
```

the build is green

could you please share your steps to reproduce the problem?

peutch:
Hi,
Starting from a fresh `git clone git@github.com:apache/openjpa.git', and importing it into my Eclipse env gives me a lot of checkstyle notifications about newline encoding and still more related to some System.(err|out) usages.
Then I just checked it using the previous `find` command.
Would this be a strange side effect of Eclipse?

Otherwise there is also another tiny commit in this MR that is more a bug, I think.

rmannibucau:
Hi @peutch , if you use eclipse checkstyle, ensure to use openjpa checkstyle configuration (it is done in the pom) otherwise you fix your IDE setup issues and not openjpa project issues.
side note: project builds fine on java 8 and 11 AFAIK

peutch:
Hi @rmannibucau, on my side project builds fine on java 8 but on 11:
```
$ java -version
openjdk version "11.0.11-ea" 2021-04-20
OpenJDK Runtime Environment (build 11.0.11-ea+8-post-Debian-1)
OpenJDK 64-Bit Server VM (build 11.0.11-ea+8-post-Debian-1, mixed mode, sharing)
```
the `mvn compile` failed with a known trouble (https://bugs.openjdk.java.net/browse/JDK-8212636):
``` 
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.0:compile (default-compile) on project openjpa-lib: Compilation failure
[ERROR] /home/patrice/local/openjpa/openjpa-lib/src/main/java/org/apache/openjpa/lib/conf/ConfigurationImpl.java:[522,18] cannot access com.sun.beans.introspect.PropertyInfo
[ERROR]   class file for com.sun.beans.introspect.PropertyInfo not found 
```

rmannibucau:
@peutch yes it needs to enforce release 8 or equivalent using java 11. Can be done with a flag of maven compiler if bothering.

peutch:
@rmannibucau it is fine thank! (both 8, 11, 14 and 17 are available for my system)
I am trying to setup a patched version of openjpa to fix my point here:
http://openjpa.208410.n2.nabble.com/please-help-with-untime-trouble-while-loading-XML-org-xml-sax-SAXException-td7593078.html
as openjpa is a dependency in many of my projects.
I will check the status and work also on the following:
https://issues.apache.org/jira/browse/OPENJPA-2729
https://issues.apache.org/jira/browse/OPENJPA-2805

struberg:
ohhh uhhh please not! We dtry to keep our history up. If we do change those lines we will see nothing with git-blame.
We also do not finally fix all that nasty tabs for the exact same reason.

struberg:
hmm, I did not find the knob for 'request change' so I closed this now to prevent getting it merged too hastily. Happy to work on your problems with you and will also check your tickets!

struberg:
I looked at the ticket and yes, we should really address that. We should not try to load XSD from some remote repositories - will look at it!

solomax:
This commit might worth cherry-picking https://github.com/apache/openjpa/pull/78/commits/2dd937f2321343783e91a952820630033bae25c5

peutch:
Sorry for this (my env is better now) and thanks for all.

PR #79, opened 2021-04-10
HerdDB profile for integration tests

eolivelli:
This is a draft, I am going to add a profile that adds HerdDB 0.20.0 dependency and start HerdDB server with docker

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1102,6 +1102,58 @@
             </properties>
         </profile>
 
+        <!-- Profile for testing with HerdDB -->
+        <profile>
+            <id>test-herddb-docker</id>
+            <activation>
+                <property>
+                    <name>test-herddb-docker</name>
+                </property>
+            </activation>
+            <dependencies>
+                <dependency>
+                    <groupId>org.herddb</groupId>
+                    <artifactId>herddb-jdbc</artifactId>
+                    <version>0.22.0</version>
+                    <scope>test</scope>
+                </dependency>
+            </dependencies>
+            <properties>
+                <connection.driver.name>org.herddb.jdbc.Driver</connection.driver.name>
+
+                <connection.url>jdbc:herddb://localhost</connection.url>
+                <connection.username>sa</connection.username>
+                <connection.password>hdb</connection.password>
+                <jdbc.DBDictionary />
+
+                <!-- DBCP overrides for testing -->
+                <dbcp.maxIdle>5</dbcp.maxIdle>
+                <dbcp.minIdle>0</dbcp.minIdle>
+
+            </properties>
+
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>io.fabric8</groupId>
+                        <artifactId>docker-maven-plugin</artifactId>
+                        <configuration>
+                            <images>
+                                <image>
+                                    <name>herddb/herddb:latest</name>
+                                    <run>
+                                        <ports>
+                                            <port>7000:7000</port>
+                                        </ports>
+                                    </run>
+                                </image>
+                            </images>
+                        </configuration>
+                    </plugin>
+                </plugins>
+            </build>
+        </profile>
+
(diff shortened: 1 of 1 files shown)


eolivelli:
@struberg @rmannibucau FYI
I am just starting...
I hope we can have something useful to test the release

pom.xml
```diff
@@ -1102,6 +1102,58 @@
...
+            <dependencies>
+                <dependency>
+                    <groupId>org.herddb</groupId>
+                    <artifactId>herddb-jdbc</artifactId>
+                    <version>0.22.0</version>
+                    <scope>test</scope>
+                </dependency>
+            </dependencies>
+            <properties>
+                <connection.driver.name>herddb.jdbc.Driver</connection.driver.name>
```
rmannibucau, line 1122: what about a plain java test since herddb is java? docker is mainly an additional test to enable to test which looks not needed here, wdyt?

struberg: For me docker is also fine. The benefit of a docker image is that you drop it and have a very well defined initial status.

eolivelli: the main problem is about adding more jars in the classpath.

I am fine with using in-memory db as well, tests will be probably faster? 

we should have both.

I would start with docker in order to not need to spend time in debugging potential problems about re-initialization of the server

rmannibucau: I would dependency:unpack the zip/stack and just launch it with a dedicated classloader (no conflict wth openjpa)
Docker is fine but don't expect it to be ran often except on the CI from time to time - this is my main point, only Mark runs docker profiles AFAIK.

eolivelli:
@rmannibucau @struberg  I am able to run the tests using HerdDB is pure "embedded" mode.
the problem is that there are many tests that are failing.

When I developed the fixes for OpenJPA I did not run all of the tests, but only a basic subset.

I have updated the patch, with `-Dtest-herddb ` you run the tests using HerdDB
`mvn clean install -Dtest-herddb`

eolivelli:
I added a few references to the docs about the ways to run HerdDB

probably many failures are due to:
- missing using of "delimiters"
- unsupported syntaxes in HerdDB (all in all it is a new dialect....)

This is a minimal test case project we have in HerdDB repository to ensure that OpenJPA is working with the basic features
https://github.com/diennea/herddb/tree/master/herddb-thirdparty/openjpa-test

rmannibucau:
We can use assume to skip unsupported tests I guess for now and flag them explicitly as "todo" maybe? (For next release to not block it)

eolivelli:
@rmannibucau @struberg I am adding assumptions in order to skip test that fail to HerdDB.

some tests fail due to assertions that do not keep into consideration that the dictionary is applying delimiters, some tests are due to some other syntax error (probably something not supported by Calcite) and some other errors fail due to some DB error that needs more investigation.

openjpa-persistence-jdbc/src/test/java/org/apache/openjpa/persistence/TestUnwrap.java
```diff
@@ -30,12 +30,22 @@
...
+import org.apache.openjpa.persistence.detachment.model.NoDetachedStateEntityPropertyAccess;
 import org.apache.openjpa.persistence.test.SingleEMFTestCase;
 
+import static org.junit.Assume.assumeFalse;
+
 public class TestUnwrap extends SingleEMFTestCase {
 
+    @Override
+    protected void setUp(Object... props) {
+        assumeFalse(this.getDBDictionary() instanceof HerdDBDictionary);
```
struberg, line 46: do you just want to skip this test for HerdDB? In which case you can simply call `setUnsupportedDatabases(HerdDBDictionary.class)`

eolivelli: @struberg thanks for the hint.
I wasn't aware of setUnsupportedDatabases 

I will rework this patch

PR #84, opened 2021-10-21
[OPENJPA-2883] 'supportsAutoAssign' is turned OFF when 'useTriggersForAutoAssign' is ON

solomax:


solomax:
Thanks for the review @rmannibucau 

1. should be improved :)
2. While I would agree with this point I see couple of possible explanations :)))

    - in this particular case user explicitly specifies `UseTriggersForAutoAssign` while `supportsAutoAssign` is auto-set in the code
    - these two can't be enabled simultaneously :((( please see the description in JIRA
    - I believe this issue is simple typo, so I tried to create very simple fix :)))

Would be happy to further improve this PR :)

solomax:
BTW fresh build always fails because integration tests for `openjpa-maven-plugin` depending on `OpenJPA Aggregate Jar` which is being built later ....
Shall the dependency be added?

PR #89, opened 2021-11-23
[OPENJPA-2881] orm* schemas are available offline

solomax:


openjpa-persistence/src/main/java/org/apache/openjpa/persistence/XMLPersistenceMetaDataParser.java
```diff
@@ -575,17 +577,25 @@ protected Object getSchemaSource() {
...
+                || (_schemaLocation != null && _schemaLocation.indexOf(ORM_XSD_1_0) > -1)) {
             ormxsd = "orm-xsd.rsrc";
             useExtendedSchema = false;
+        } else if (XMLVersionParser.VERSION_2_0.equals(_ormVersion)
+                || (_schemaLocation != null && _schemaLocation.indexOf(ORM_XSD_2_0) > -1)) {
+            ormxsd = "orm_2_0-xsd.rsrc";
+        } else if (XMLVersionParser.VERSION_2_1.equals(_ormVersion)
+                || (_schemaLocation != null && _schemaLocation.indexOf(ORM_XSD_2_1) > -1)) {
+            ormxsd = "orm_2_1.xsd.rsrc";
+            useExtendedSchema = false;
```
solomax, line 590: I'm not sure about this

openjpa-persistence/src/main/java/org/apache/openjpa/persistence/XMLPersistenceMetaDataParser.java
```diff
@@ -575,17 +577,25 @@ protected Object getSchemaSource() {
...
+        } else if (XMLVersionParser.VERSION_2_2.equals(_ormVersion)
+                || (_schemaLocation != null && _schemaLocation.indexOf(ORM_XSD_2_2) > -1)) {
+            ormxsd = "orm_2_2.xsd.rsrc";
+            useExtendedSchema = false;
```
solomax, line 594: And this

openjpa-persistence/src/main/java/org/apache/openjpa/persistence/XMLPersistenceMetaDataParser.java
```diff
@@ -575,17 +577,25 @@ protected Object getSchemaSource() {
         boolean useExtendedSchema = true;
```
solomax, line 577: multiple tests fails if orm.2.2 is used by default :(

openjpa-persistence/src/main/java/org/apache/openjpa/persistence/XMLPersistenceMetaDataParser.java
rmannibucau: I'd put it true to be able to use openjpa namespace, any reason to not do?
Also think it is worth some tests otherwise there is no real code change/feature ;)

solomax: @rmannibucau IMO the main PRO is to be able to work completely offline
I have no idea how to write test like this :(

rmannibucau: use the JVM HttpServer, mock an xsd you use in the test persistence.xml and ensure the server handler is not hit during an emf->em creation (+close), will be sufficient.

solomax: @rmannibucau  I hope these tests will be OK :)

solomax: @rmannibucau is this one OK to be merged? :)

PR #90, opened 2021-11-30
[OPENJPA-2891] allow configuration of @Generated annotation.

bmarwell:
Issue: https://issues.apache.org/jira/browse/OPENJPA-2891

Question: The annotation dependency is only necessary if compiled with Java9+, but this project cannot be compiled with Java9+ at the moment. I can remove it if you would like me to.

This implementation uses a String, because this makes the parameters more extensible. You can add other parameters later. `forced` is used instead of `true` because it will be added in every situation. `true` and `null` are interchangable in this implementation.

Documentation needs to be updated as well.

Also, this is my first OpenJPA commit. Please review the code style.

Logic: 
* If nothing is set, try to see if the annotation is on the classpath and we are on Java6+.
* If `-Aopenjpa.addGeneratedAnnotation=false` is set, the annotation is never added.
* If `-Aopenjpa.addGeneratedAnnotation=force` is set, the annotation is always added (regardless of classpath availability and/or Java version).

### DIFF
--- a/openjpa-persistence/pom.xml
+++ b/openjpa-persistence/pom.xml
@@ -60,5 +60,10 @@
             <version>4.2.0</version>
             <scope>provided</scope>
         </dependency>
+        <!-- annotation javax.annotation.Generated -->
+        <dependency>
+            <groupId>javax.annotation</groupId>
+            <artifactId>javax.annotation-api</artifactId>
+        </dependency>
     </dependencies>
 </project>
--- a/openjpa-persistence/src/main/java/org/apache/openjpa/persistence/meta/AnnotationProcessor6.java
+++ b/openjpa-persistence/src/main/java/org/apache/openjpa/persistence/meta/AnnotationProcessor6.java
@@ -99,7 +99,8 @@ import org.apache.openjpa.persistence.util.SourceCode;
                     "openjpa.source",
                     "openjpa.naming",
                     "openjpa.header",
-                    "openjpa.metamodel"
+                    "openjpa.metamodel",
+                    "openjpa.addGeneratedAnnotation"
                   })
 
 public class AnnotationProcessor6 extends AbstractProcessor {
@@ -111,6 +112,9 @@ public class AnnotationProcessor6 extends AbstractProcessor {
     private boolean active;
     private static Localizer _loc =  Localizer.forPackage(AnnotationProcessor6.class);
     private SourceVersion supportedSourceVersion;
+    private String addGeneratedOption;
+    private Class<?> generatedAnnotation;
+    private Date generationDate;
 
     /**
      * Category of members as per JPA 2.0 type system.
@@ -213,6 +217,8 @@ public class AnnotationProcessor6 extends AbstractProcessor {
         setNamingPolicy();
         setHeader();
         handler = new SourceAnnotationHandler(processingEnv, logger);
+        setAddGeneratedAnnotation();
+        this.generationDate = new Date();
     }
 
     /**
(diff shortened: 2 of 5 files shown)


openjpa-persistence/src/main/java/org/apache/openjpa/persistence/meta/AnnotationProcessor6.java
```diff
@@ -311,10 +314,26 @@ private void annotate(SourceCode source, String originalClass) {
...
+
+        if ("force".equals(this.addGeneratedOption)) {
+            cls.addAnnotation(javax.annotation.Generated.class.getName())
+                    .addArgument("value", this.getClass().getName())
+                    .addArgument("date", new Date().toString());
+        }
+
+        try {
+            // only add the annotation if it is on the classpath for Java 6+.
+            Class<?> generatedAnnotation = Class.forName("javax.annotation.Generated", false, null);
```
rmannibucau, line 329: guess we want to do it in init and not each time

openjpa-persistence/src/main/java/org/apache/openjpa/persistence/meta/AnnotationProcessor6.java
```diff
@@ -311,10 +314,26 @@ private void annotate(SourceCode source, String originalClass) {
...
-            .addArgument("value", this.getClass().getName())
-            .addArgument("date", new Date().toString());
+        if ("false".equals(this.addGeneratedOption)) {
+            return;
+        }
+
+        if ("force".equals(this.addGeneratedOption)) {
+            cls.addAnnotation(javax.annotation.Generated.class.getName())
+                    .addArgument("value", this.getClass().getName())
+                    .addArgument("date", new Date().toString());
```
rmannibucau, line 324: know it was "before" but can be worth storing the date in init if needed (once the forName moved there it will be easy to know)

openjpa-persistence/src/main/java/org/apache/openjpa/persistence/meta/AnnotationProcessor6.java
```diff
@@ -379,6 +398,19 @@ private void setHeader() {
         }
     }
 
+    private void setAddGeneratedAnnotation() {
+        String addGeneratedOption = getOptionValue("openjpa.addGeneratedAnnotation");
```
rmannibucau, line 402: `this.addGeneratedOption = getOptionValue("openjpa.addGeneratedAnnotation")` ?

openjpa-persistence/pom.xml
```diff
@@ -60,5 +60,10 @@
             <version>4.2.0</version>
             <scope>provided</scope>
         </dependency>
+        <!-- annotation javax.annotation.Generated -->
+        <dependency>
+            <groupId>javax.annotation</groupId>
```
rmannibucau, line 65: provided?

rmannibucau (changes requested):

openjpa-persistence/src/main/java/org/apache/openjpa/persistence/meta/AnnotationProcessor6.java
bmarwell: Will do!

rmannibucau (approved):
think addGeneratedOption can be worth a default (auto or whatever) to use a switch instead of chain of "if" (which would avoid the current double if check in terms of pure code path ;)) but code looks more than ok too so will not be picky since it does the job as expected for me

bmarwell:
> think addGeneratedOption can be worth a default (auto or whatever)

Can do, that's not much effort and sounds like a sensible option.

PR #92, opened 2021-12-17
[OPENJPA-2896] automatic module name is set for bundles

solomax:


openjpa-all/pom.xml
```diff
@@ -110,6 +110,7 @@
                                         <Specification-Title>JSR-338 Java Persistence</Specification-Title>
                                         <Specification-Vendor>ORACLE, Inc.</Specification-Vendor>
                                         <Specification-Version>2.2</Specification-Version>
+                                        <Automatic-Module-Name>org.apache.openjpa.all</Automatic-Module-Name>
```
rmannibucau, line 113: isnt it a broken pattern to not have the same name for standard dep and bundle one? should be replaceable and it is no more?

Maybe leave the bundle away from JPMS or only enable JPMS for it but not both, wdyt?

solomax: I'm not sure I understand what is wrong :(
I believe every module need it's own unique name :))

I've added automatic module name to `openjpa-all` for consistency

I try to migrate my main project to JPMS and fix all 
```
[WARNING] * Required filename-based automodules detected: [openjpa-3.2.1-SNAPSHOT.jar]. Please don't publish this project to a public artifact repository! *
```

I can leave `openjpa-all`as-is in this PR, would it be better? :)

rmannibucau: > I believe every module need it's own unique name :))

Right but when you do an all on one you dont create a new module but you merge N modules.
Issue is on consumer side, you depend on org.apache.openjpa.kernel, with this strategy your module-info will depend on openjpa.all or openjpa.kernel (ie the deployment is bound to the build now) which is quite unlikely so it is better to have stable names for the same packages.
This is why I think we should choose to enable JPMS for only the fatjar or only independent modules but not both.

mattrpav: Agreed

solomax: right now I  have dependency on openjpa-3.2.1-SNAPSHOT.jar (which is fatjar as well)
and I have no option to add `org.apache.openjpa.kernel`
only `openjpa` with WARN :(

what am I doing wrong?

rmannibucau: openjpa is a modular project, you can use individual jar. It is common to use the fat bundle but it leads to break JPMS as explained so we should choose if we provide JPMS for modules or just for the fatjar (or nobody ;)). What we must not do is to provide not stable names regarding packages (ignore the .jar, only packages are important and your module-info should be stable)

solomax: From what I can see some project do have "Automatic Module name" for both fat-jars and regular jars
for example Apache Tika

I would expect most of users will use fat-jar-dependency

Not sure what should be my next steps :(

rmannibucau: Yes, some do and break users (unintentionally from the ones I saw - mainly a "put automatic names everywhere" without caring about consumer side :().
The topic is interesting and joins another one: should our fatjar be the recommended one or should we promote the modular approach? Once this question answered we just put automatic names - or real module-info - in these modules.
One approach is to get stats from public github repos using openjpa.
I tend to think like you but without figures it is quite hypothetic IMHO.
Wdyt?

solomax: Approx. numbers available here:

https://mvnrepository.com/artifact/org.apache.openjpa/openjpa
https://mvnrepository.com/artifact/org.apache.openjpa/openjpa-persistence-jdbc
https://mvnrepository.com/artifact/org.apache.openjpa/openjpa-kernel

rmannibucau: Well I would really use github cause mvnrepo just indexes central and very few projects are released to central (guess central download stats are an alternative approach which can be better). We are at 40% for the fatjar, 11 for the karaf feature and rest is distributed <= 7% between other artifacts so our assumption was right, let's name the module only for the fatjar :).

solomax: Should be addressed :)

rmannibucau (approved):
Okish for me, points to check before hitting merge:

1. do we want `org.apache.openjpa.jakarta`? assume we move from javax to jakarta in the code, we would keep `org.apache.openjpa` so for me we should use the same name (both can't be in the same classloader anyway)
2. why `org.apache.openjpa` and `org.apache.openjpa.all`, due to the reasoning we had on this PR I think a single one is valid.

solomax:
Shall I drop `Auto-module` for `openjpa-all` artifact?
not using it since it is too fat :)))

rmannibucau:
@solomax let's just have openjpa one then ;)

PR #93, opened 2022-02-21
OPENJPA-2900 javax.xml.bind must be reloacted to jakarta namespace

jeanouii:
It also cleans up unused relocations and adds some excludes for jakarta exceptions.

### DIFF
--- a/openjpa/pom.xml
+++ b/openjpa/pom.xml
@@ -129,48 +129,19 @@
                             </artifactSet>
                             <relocations>
                                 <relocation>
-                                    <pattern>javax.annotation.security</pattern>
-                                    <shadedPattern>jakarta.annotation.security</shadedPattern>
+                                    <pattern>javax.annotation</pattern>
+                                    <shadedPattern>jakarta.annotation</shadedPattern>
+                                    <excludes>
+                                        <exclude>javax.annotation.process.**</exclude>
+                                    </excludes>
                                 </relocation>
                                 <relocation>
-                                    <pattern>javax.annotation.Generated</pattern>
-                                    <shadedPattern>jakarta.annotation.Generated</shadedPattern>
+                                    <pattern>javax.jms</pattern>
+                                    <shadedPattern>jakarta.jms</shadedPattern>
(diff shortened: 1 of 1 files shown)


openjpa/pom.xml
```diff
@@ -129,48 +129,19 @@
...
-                                <relocation>
-                                    <pattern>javax.annotation.Resource</pattern>
-                                    <shadedPattern>jakarta.annotation.Resource</shadedPattern>
-                                </relocation>
-                                <relocation>
-                                    <pattern>javax.annotation.Resources</pattern>
-                                    <shadedPattern>jakarta.annotation.Resources</shadedPattern>
-                                </relocation>
-                                <relocation>
-                                    <pattern>javax.el</pattern>
```
solomax, line 156: these seems to be lost ...

openjpa/pom.xml
```diff
@@ -179,34 +150,18 @@
...
                                     <shadedPattern>jakarta.validation</shadedPattern>
                                 </relocation>
                                 <relocation>
-                                    <pattern>javax.persistence</pattern>
-                                    <shadedPattern>jakarta.persistence</shadedPattern>
+                                    <pattern>javax.xml.bind</pattern>
+                                    <shadedPattern>jakarta.xml.bind</shadedPattern>
                                 </relocation>
-                                <relocation>
-                                    <pattern>javax.faces</pattern>
```
solomax, line 192: and also these

openjpa/pom.xml
rmannibucau: not used ;)

but guess it is worth a test ran after shade goal execution to validate the produced jar does not contain any javax.xxx which is unexpected (using asm since we have it in the cp)

jeanouii: Indeed and on purpose. There is no use of EL in OpenJPA. So no need to have them in my opinion.

openjpa/pom.xml
jeanouii: Same here for Faces

PR #101, opened 2022-09-15
Log4j2 update to last version (free of vulnerability such as Log4Shell)

s4iko:
Hi,
The fact is that old version of Log4j2 are sometimes filtered by proxies resulting in a build failed due to missing dependency (that alos impact other project depending on this one, such as Camel).
And in a more global reflection, I think it is better to have a "clean" log4j2 version ;)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -99,7 +99,7 @@
 
         <xbean.version>4.20</xbean.version>
         <bval.version>1.1.2</bval.version>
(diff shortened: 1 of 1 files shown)


rmannibucau (approved):

solomax (approved):

solomax:
Thanks for the contribution @s4iko !

PR #111, opened 2023-06-05
3.2.2 fix openjpa 2814

gjevardat:
I reopened [OPENJPA-2814](https://issues.apache.org/jira/browse/OPENJPA-2814) because the proposed fix was causing some exceptions.

My initial fix is fixing properly the memory leak and is in production since few years already so I propose to use this one.

### DIFF
--- a/openjpa-jdbc/src/main/java/org/apache/openjpa/jdbc/schema/Constraint.java
+++ b/openjpa-jdbc/src/main/java/org/apache/openjpa/jdbc/schema/Constraint.java
@@ -14,7 +14,7 @@
  * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
  * KIND, either express or implied.  See the License for the
  * specific language governing permissions and limitations
- * under the License.
+ * under the License.    
  */
 package org.apache.openjpa.jdbc.schema;
 
@@ -27,8 +27,10 @@ import org.apache.openjpa.jdbc.identifier.QualifiedDBIdentifier;
  *
  * @author Abe White
  */
-public abstract class Constraint extends ReferenceCounter {
-    private static final long serialVersionUID = 1L;
+@SuppressWarnings("serial")
+public abstract class Constraint
+    extends ReferenceCounter {
+
     private DBIdentifier _name = DBIdentifier.NULL;
     private QualifiedDBIdentifier _fullPath = null;
     private Table _table = null;
@@ -50,7 +52,6 @@ public abstract class Constraint extends ReferenceCounter {
      * @param table the local table of the constraint
      * @deprecated
      */
-    @Deprecated
     Constraint(String name, Table table) {
         this(DBIdentifier.newConstant(name), table);
     }
@@ -83,7 +84,6 @@ public abstract class Constraint extends ReferenceCounter {
      * Return the column's table name.
      * @deprecated
      */
-    @Deprecated
     public String getTableName() {
         return getTableIdentifier().getName();
     }
@@ -97,7 +97,6 @@ public abstract class Constraint extends ReferenceCounter {
      * columns whose table object is not set.
      * @deprecated
      */
-    @Deprecated
     public void setTableName(String name) {
         setTableIdentifier(DBIdentifier.newTable(name));
     }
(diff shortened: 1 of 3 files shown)


openjpa-jdbc/src/main/java/org/apache/openjpa/jdbc/schema/Constraint.java
```diff
@@ -250,31 +241,48 @@ public String toString() {
         return "<" + name.toLowerCase() + ">";
     }
 
-    @Override
```
solomax, line 253: IMO previous version of hashcode+equals was more readable

also it not clear why to change the order :)

openjpa-jdbc/src/main/java/org/apache/openjpa/jdbc/schema/ForeignKey.java
```diff
@@ -40,8 +40,9 @@
  *
  * @author Abe White
  */
-public class ForeignKey extends Constraint {
-    private static final long serialVersionUID = 1L;
```
solomax, line 44: IMo this code:
```
public class ForeignKey extends Constraint {
    private static final long serialVersionUID = 1L;
```

need to be restored

openjpa-jdbc/src/main/java/org/apache/openjpa/jdbc/schema/ForeignKey.java
```diff
@@ -183,45 +182,45 @@ public boolean isLogical() {
...
+        if (_autoAssign != null) 
+            return _autoAssign.booleanValue();
 
         Column[] cols = getPrimaryKeyColumns();
         if (cols.length == 0) {
             _autoAssign = Boolean.FALSE;
             return false;
         }
 
-        for (Column column : cols) {
```
solomax, line 210: Could you please clarify why `for-in` loops were replaced with `plain-for` loops in this PR?

openjpa-jdbc/src/main/java/org/apache/openjpa/jdbc/schema/ForeignKey.java
```diff
@@ -810,58 +795,58 @@ public DBIdentifier loadIdentifierFromDB(DBDictionary dbdict, Connection conn) {
...
                         if (log.isTraceEnabled()) {
                             log.trace(_loc.get("fk-column-mismatch"));
                         }
                     }
                     for (int j = 0; j < locCols.length; j++) {
-                        if (!fkTemp.containsColumn(
-                                localtable.getColumn(locCols[j].getIdentifier()))) {
-                            fkTemp.join(localtable.getColumn(locCols[j].getIdentifier()),
-                                    pkTable.getColumn(pkCols[j].getIdentifier()));
+                        if( ! fkTemp.containsColumn(
```
solomax, line 835: `if (...)` is used in the code

not `if( ...)`

could you please revert this change?

openjpa-jdbc/src/main/java/org/apache/openjpa/jdbc/schema/ForeignKey.java
```diff
@@ -920,72 +905,60 @@ private static Column[] createKeyColumns(ForeignKey fk) {
         return new Column[] { fkCol, pkCol };
     }
 
-    @Override
-    public boolean equals(Object o) {
```
solomax, line 924: same comment to hashcode/equals

solomax (changes requested):

openjpa-jdbc/src/main/java/org/apache/openjpa/jdbc/schema/Constraint.java
gjevardat: I used an auto generated eclipse equals and hashcode. I could mimic the style of the original one indeed. The important thing for me is the fields that are included in the equality and hashcode testing to avoid the mem leak !

openjpa-jdbc/src/main/java/org/apache/openjpa/jdbc/schema/ForeignKey.java
gjevardat: Yes see my general comment, will update

PR #114, opened 2023-09-27
[OPENJPA-2915] properties of type Duration can be set as expected

solomax:


rmannibucau (approved):

rmannibucau:
@solomax don't think so, we can do it lazily if there is any request

PR #123, opened 2025-02-03
[OPENJPA-2933] Implements new date JPA 3.1 JPQL functions and equivalent Criteria API

cristof:
Implemented LOCAL DATE, LOCAL TIME, LOCAL DATETIME functions and criteria api equivalents
Implemented EXTRACT functions
Updated documentation
Fixed corner case of round function on postgresql (got the error when testing updated master on derbydb, h2, mariadb (11), mysql (8 and 9) and postgresql (16 and 17).
 Please, don't hesitate to ask fixes (pretty sure it needs some).

### DIFF
--- a/openjpa-kernel/src/main/java/org/apache/openjpa/kernel/exps/ExtractDateTimeField.java
+++ b/openjpa-kernel/src/main/java/org/apache/openjpa/kernel/exps/ExtractDateTimeField.java
@@ -1,0 +1,100 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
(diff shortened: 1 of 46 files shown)


openjpa-kernel/src/main/java/org/apache/openjpa/kernel/exps/ExtractDateTimeField.java
```diff
@@ -0,0 +1,100 @@
...
+        } else if (Temporal.class.isAssignableFrom(r.getClass())) {
+            t = (Temporal) r;
+        }
+        if (t == null) {
+            throw new IllegalArgumentException();
+        }
+        switch (_field) {
+            case QUARTER:
+                int month = t.get(ChronoField.MONTH_OF_YEAR);
+                return (int) Math.round(Math.ceil(month/3f));
```
solomax, line 83:
```suggestion
                return Math.round(month / 3f);
```
should be enough here :)

solomax (commented):
I would ask for a minor change

The PR is absolutely fantastic! Thanks for you work! :))

openjpa-kernel/src/main/java/org/apache/openjpa/kernel/exps/ExtractDateTimeField.java
cristof: Actually, your solution gives wrong results for January(1) - gives 0, April(4) - gives 1, July (7) - gives 3, and October (10) - gives 3, because rounds 0.3 to 0 and so on.

solomax: My bad :(
You are right :)

The only redundant part is `(int)` :))

PR #124, opened 2025-02-15
OPENJPA-2930 Implements UUID and GenerationType.UUID strategy

cristof:
Added UUID support, using binary storage when the pair database-jdbc driver supports it, and string storage otherwise.
Binary storage is implemented as an DbDictionary property, defaulting to false, and true for PostgreSQL, H2 and HSQL databases.

Generated UUID value is a type 4 uuid, and storage of string representation is made, per spec, using its canonical representation.

Tested new features agains Postgresql 13 to 17, H2 1 and H2 2, current UUID capable HSQLDB and previous UUID not capable HSQLDB, MariaDB lts and latest, MySQL lts and latest. Could not test against Oracle.

Please be kind with this one. I may have missed a lot.


openjpa-kernel/src/main/java/org/apache/openjpa/abstractstore/AbstractStoreManager.java
```diff
@@ -42,6 +42,7 @@
 import org.apache.openjpa.util.ApplicationIds;
 import org.apache.openjpa.util.Id;
 import org.apache.openjpa.util.ImplHelper;
+import org.apache.openjpa.util.UuidId;
```
solomax, line 45: Unused import?

openjpa-kernel/src/main/java/org/apache/openjpa/meta/JavaTypes.java
```diff
@@ -417,6 +421,11 @@ public static Object convert(Object val, int typeCode) {
                 if (val instanceof String)
                     return Short.valueOf(val.toString());
                 return val;
+            case UUID_OBJ:
+                if (val instanceof String) {
+                    return UUID.fromString(val.toString());
```
solomax, line 426:
```suggestion
                    return UUID.fromString((String)val);
```
?

openjpa-kernel/src/main/java/org/apache/openjpa/util/UuidId.java
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    @Override
+    protected int idHash() {
+        return (_key == null) ? 0 : _key.hashCode();
+    }
+
+    @Override
+    protected boolean idEquals(OpenJPAId o) {
+        Object key = ((UuidId) o)._key;
```
solomax, line 69: this one might throw :(

I would write something like this:

`return (_key == null && o == null) || (_key != null && _key.equals(o)`

:)

solomax (commented):
As always great PR :))
Thanks a million! :)

openjpa-kernel/src/main/java/org/apache/openjpa/util/UuidId.java
solomax: My bad :(

I wrote something stupid

for java 11 it should be

```
    if (!(o instanceof UuidId)) {
        return false
    }
    UUID key = ((UuidId) o)._key;
    return (_key == null) ? key == null : _key.equals(key);
```

openjpa-jdbc/src/main/java/org/apache/openjpa/jdbc/sql/DBDictionary.java
```diff
@@ -1956,6 +1967,15 @@ public String getTypeName(Column col) {
         if (col.isAutoAssigned() && autoAssignTypeName != null)
             return appendSize(col, autoAssignTypeName);
 
+        if (col.getJavaType() == JavaTypes.UUID_OBJ) {
+            if (supportsUuidType) 
+                return appendSize(col, uuidTypeName);
+            else {
+                //col.setSize(36);
```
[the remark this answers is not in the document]
cristof: Cleaned

openjpa-kernel/src/main/java/org/apache/openjpa/meta/JavaTypes.java
cristof: I used ```val.toString()``` because of how other cases use it (such as LONG, INT etc.). Exchanged for (String) val as suggested and tested again on hsql, h2, h2-2, pg-13 to pg-17, mysql lts and latest, mariadb lts and latest, and oracle XE 21c.

openjpa-kernel/src/main/java/org/apache/openjpa/util/UuidId.java
cristof: Fixed as suggested.

PR #129, opened 2025-03-17
OPENJPA-2925 - Removed stalled eager result

cristof:
Removed cached result just before loading eager joins. It should not impact direct eager loading. 

The cached result ignored the difference between an eager loaded entity (Cat->human) and a call to a parent who also have the same eager loaded entity (Cat->parent(Cat)->human), in the case getting the last loaded value from (Cat->parent->cats).

I tried to remove this cache in another place, but it broke multiple eager loading tests (that tested for a limit of db calls). Didn't check if there is a larger impact, though.


solomax (approved):
LGTM!

solomax:
Thanks a lot @cristof , Thanks for your patch!

PR #133, opened 2025-06-22
[OPENJPA-2937] Merging previously unsaved entity with generated UUID property fixed

cristof:
When an entity containing UUID generated fields where merged without being previously persisted, OpenJPA throwed an InternalException because AttachStrategy was not prepared for this type of field.
Added tests for this scenario and for detaching and getReference.


solomax (approved):
LGTM :)

solomax:
BTW maybe you know what can be wrong with our GH builds? :))

clambertus:
Your test database is missing a table: Caused by: org.apache.openjpa.lib.jdbc.ReportingSQLException: Table/View 'OPENJPA_SEQUENCE_TABLE' does not exist. {stmnt 1594403526 DELETE FROM OPENJPA_SEQUENCE_TABLE} [code=30000, state=42X05]

PR #135, opened 2025-07-04
[OPENJPA-2941] Switching CI to H2 DB (with DB driver update and test fix)

solomax:


cristof:
Wouldn't it better to change the profile to test-h2-2 instead of changing the test-h2 profile to use h2-2.x version? In my forked OPENJPA-2940 branch, I 've found some other issues with the h2 and h2-2 tests.

solomax:
This https://ci-builds.apache.org/job/OpenJPA/job/openjpa-pr/ build passes on this branch
I'll re-test with 1.x :)
But according to https://mvnrepository.com/artifact/com.h2database/h2 all 1.x versions are vulnerable
And latest is fron 2019 .....

solomax:
I'm currently investigating :)

Current results are:

- my local build on derby is green
- jenkins derby build is red
- my local build on h2 1.x is red (need to be re-tested)
- my local build on h2 2.x is green (in branch)
- gh build on h2 2.x is red (on branch)
- jenkins build h2 2.x is green (on branch)

I need to re-test above and clean-up the pr

I would liks to keep derby as default since it works locally :)

PR #149, opened 2026-05-22
[OPENJPA-3002] Fix: @Index(columnNames) at field level is now correctly applied

cwegener-79:
### Problem

The OpenJPA-specific `@Index` annotation (`org.apache.openjpa.persistence.jdbc.Index`) supports a `columnNames` attribute that allows explicitly defining which columns should be included in a database index when the annotation is placed on a field or method. However, these column names were **silently ignored** during annotation parsing.

**Root cause:** `AnnotationPersistenceMappingParser.parseIndex(MappingInfo, Index)` called an internal overload passing only `name`, `enabled`, and `unique` — `idx.columnNames()` was never forwarded:

```java
// Before — columnNames dropped on the floor:
private void parseIndex(MappingInfo info, Index idx) {
    parseIndex(info, idx.name(), idx.enabled(), idx.unique());
}
```

As a result, any entity using `@Index(columnNames = {"COL_A", "COL_B"})` at the field level would get a schema index with no explicitly defined columns.

### Fix

- `parseIndex(MappingInfo, Index)` now passes `idx.columnNames()` to a new overload.
- New method `parseIndex(MappingInfo, String, boolean, boolean, String[])` creates `Column` objects from the provided names and adds them to the `Index` schema object.
- The existing `protected` 4-parameter overload delegates to the new method with `null` column names, preserving backwards compatibility for subclasses.

### Testing

Added a regression test that verifies the fix end-to-end:

- **`EntityWithIndexColumnNames`** — test entity with `@Index(name="idx_col_a_b", columnNames={"COL_A","COL_B"})` on a field.
- **`TestIndexColumnNames#testFieldIndexColumnNamesAreApplied`** — asserts that after full mapping resolution the schema index contains both explicitly named columns. The test fails without the fix (`getColumns().length == 0`) and passes with it.

### Changed Files

| File | Change |
|---|---|
| `AnnotationPersistenceMappingParser.java` | Bug fix + whitespace/indentation cleanup |
| `EntityWithIndexColumnNames.java` | New test entity |
| `TestIndexColumnNames.java` | New regression test |

### DIFF
--- a/openjpa-persistence-jdbc/src/main/java/org/apache/openjpa/persistence/jdbc/AnnotationPersistenceMappingParser.java
+++ b/openjpa-persistence-jdbc/src/main/java/org/apache/openjpa/persistence/jdbc/AnnotationPersistenceMappingParser.java
@@ -1063,7 +1063,7 @@ public class AnnotationPersistenceMappingParser
      * Parse the given index.
      */
     private void parseIndex(MappingInfo info, Index idx) {
-    parseIndex(info, idx.name(), idx.enabled(), idx.unique(), idx.columnNames());
+        parseIndex(info, idx.name(), idx.enabled(), idx.unique(), idx.columnNames());
     }
 
     /**
@@ -1090,11 +1090,12 @@ public class AnnotationPersistenceMappingParser
             idx.setIdentifier(DBIdentifier.newConstraint(name, delimit()));
         idx.setUnique(unique);
 
-        if (columnNames != null) for (String columnName : columnNames) {
-            org.apache.openjpa.jdbc.schema.Column column =
-                    new org.apache.openjpa.jdbc.schema.Column();
-            column.setIdentifier(DBIdentifier.newColumn(columnName, delimit()));
-            idx.addColumn(column);
+        if (columnNames != null) {
+            for (String columnName : columnNames) {
+                Column column = new Column();
+                column.setIdentifier(DBIdentifier.newColumn(columnName, delimit()));
+                idx.addColumn(column);
+            }
         }
 
         info.setIndex(idx);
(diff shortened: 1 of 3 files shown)


cwegener-79:
Just a quick nudge on this PR 🙂
I know @Index is not part of standard JPA, but I could make good use of this fix in a legacy project.
Please let me know if there’s anything I can do to help move this forward.

openjpa-persistence-jdbc/src/main/java/org/apache/openjpa/persistence/jdbc/AnnotationPersistenceMappingParser.java
```diff
@@ -918,40 +918,56 @@ private int toForeignKeyAction(ForeignKeyAction action) {
...
         }
 
         org.apache.openjpa.jdbc.schema.Index idx =
-            new org.apache.openjpa.jdbc.schema.Index();
+                new org.apache.openjpa.jdbc.schema.Index();
         if (!StringUtil.isEmpty(name))
             idx.setIdentifier(DBIdentifier.newConstraint(name, delimit()));
         idx.setUnique(unique);
+
+        if (columnNames != null) for (String columnName : columnNames) {
```
solomax:
```suggestion
        if (columnNames != null) {
            for (String columnName : columnNames) {
```

could you please change this `if-for` to have block for `if` with curly braces? (IMO it will be more readable)

solomax (changes requested):
Could you please remove all whitespaces-only changes to `AnnotationPersistenceMappingParser`

Thanks in advance :)

openjpa-persistence-jdbc/src/main/java/org/apache/openjpa/persistence/jdbc/AnnotationPersistenceMappingParser.java
cwegener-79: You’re right, using curly braces makes it more readable.

I’ve removed the whitespace-only changes in AnnotationPersistenceMappingParser, restored the accidental indentation changes. By the way I updated my commit signature as well. Now it's signed and verified. :-)

Please let me know if there’s anything else I can do for this fix.

solomax:
@cwegener-79 could you please rebase this one? we just have merged huge new functionality (this is why merging was delayed)

cwegener-79:
> @cwegener-79 could you please rebase this one? we just have merged huge new functionality (this is why merging was delayed)

No Problem, I've just rebased it.

solomax:
Thanks for the contribution!

PR #152, opened 2026-08-20
[OPENJPA-2987] Restore externalized-parameter assertions

rzo1:
Follow-up on https://github.com/apache/openjpa/pull/144#discussion_r3683006222 (OPENJPA-2987).

### Answer to the review question

Externalized-parameter detection was **not** removed. `PreparedQueryImpl.extractSelectExecutor()` still calls `isUsingExternalizedParameter(...)` and excludes such queries from the prepared query cache. With trace logging enabled:

```
Query "select b from Book b where b.token=:token" is removed from cache excluded permanently.
... is not cached because some parameterized field values are externalized.
```

What broke was only the *test's access* to the expressions: JPA 3.2 requires a mutable `getResultList()`, so `org.apache.openjpa.persistence.QueryImpl.getResultList()` now returns `new ArrayList<>(delegate)` and the `(ResultList) result` cast in the old `getExpressions()` helper no longer works. The cache itself is unaffected, since `postExecute()` still receives the raw `ResultList`.

### Change

`TestExternalizedParameter` gets its original assertions back:

- `getExpressions(...)` helper restored
- `assertFalse(...)` / `assertTrue(isUsingExternalizedParameter(exps[0]))` restored in place of the no-op `assertNotNull(getResultList())`
- the `ResultList` is obtained by executing the kernel query directly (`((QueryImpl<?>) em.createQuery(jpql)).getDelegate().execute(params)`) instead of through the JPA facade

### Test

`mvn -pl openjpa-persistence-jdbc -Dtest=TestExternalizedParameter test` -> Tests run: 3, Failures: 0, Errors: 0.

### Note

While looking for a cache-level assertion instead: `PreparedQueryCacheImpl.isExcluded(id)` only matches user-supplied exclusion *patterns*. Per-query exclusions are stored in `_uncachables` and are only observable via `isCachable(id) == FALSE`, with no accessor for the exclusion reason. Since every `Book` query is uncachable anyway (eager `@ManyToMany` -> multiple SQL statements), that route cannot distinguish an externalized-parameter exclusion from any other, so the kernel-level assertion is the one that actually discriminates.

### DIFF
--- a/openjpa-persistence-jdbc/src/test/java/org/apache/openjpa/persistence/jdbc/sqlcache/TestExternalizedParameter.java
+++ b/openjpa-persistence-jdbc/src/test/java/org/apache/openjpa/persistence/jdbc/sqlcache/TestExternalizedParameter.java
@@ -18,7 +18,9 @@
  */
 package org.apache.openjpa.persistence.jdbc.sqlcache;
 
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.Properties;
 
 import jakarta.persistence.EntityManager;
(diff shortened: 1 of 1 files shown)


solomax (approved):
this was fast :)

cristof (approved):

PR #155, opened 2026-08-20
[OPENJPA-2978] Fix SchemaManagerImpl nits and unsupported-validation reporting

rzo1:
Cleans up the reported nits in `SchemaManagerImpl` (stray double semicolon, redundant `(Exception)` cast, javadoc typo) and stops `validate()` from wrapping an `UnsupportedOperationException` into `SchemaValidationException("Schema could not be validated: null")` — a store that does not implement validation now reports a missing capability, consistently with `create()`, `drop()` and `truncate()`. The message-less `UnsupportedOperationException`s in `AbstractBrokerFactory` got real messages, and a new `TestSchemaManagerImpl` covers both the unsupported and the real-failure path.

The `Strucuture`/`Struture` typos from the review comment are already fixed on master (commit cbb19f777, "[OPENJPA-2940] Fixes typos in BrokerFactory"), so no rename was needed.

### DIFF
--- a/openjpa-kernel/src/main/java/org/apache/openjpa/kernel/AbstractBrokerFactory.java
+++ b/openjpa-kernel/src/main/java/org/apache/openjpa/kernel/AbstractBrokerFactory.java
@@ -475,22 +475,22 @@ public abstract class AbstractBrokerFactory implements BrokerFactory {
     
     @Override
     public void createPersistenceStructure(boolean createSchemas) {
-    	throw new UnsupportedOperationException();
+    	throw new UnsupportedOperationException("This BrokerFactory does not implement schema creation.");
     }
     
     @Override
     public void dropPersistenceStructure(boolean dropSchemas) {
-    	throw new UnsupportedOperationException();
+    	throw new UnsupportedOperationException("This BrokerFactory does not implement schema dropping.");
     }
     
     @Override
     public void validatePersistenceStructure() throws Exception {
-    	throw new UnsupportedOperationException();
+    	throw new UnsupportedOperationException("This BrokerFactory does not implement schema validation.");
     }
     
     @Override
     public void truncateData() {
-    	throw new UnsupportedOperationException();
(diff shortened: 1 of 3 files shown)


openjpa-kernel/src/main/java/org/apache/openjpa/kernel/AbstractBrokerFactory.java
```diff
@@ -475,22 +475,22 @@ public void unlock() {
     
     @Override
     public void createPersistenceStructure(boolean createSchemas) {
-    	throw new UnsupportedOperationException();
+    	throw new UnsupportedOperationException("This BrokerFactory does not implement schema creation.");
```
solomax, line 478: this class is `abstract` maybe these methods need to be just removed?

rzo1: These can't just be removed: `AbstractStoreBrokerFactory` (openjpa-kernel/src/main/java/org/apache/openjpa/abstractstore/AbstractStoreBrokerFactory.java:51) is a concrete main-code subclass that implements none of the four, as is the `BrokerFactory` stub in `TestPersistenceManagerFactoryImpl`, so both would stop compiling and we'd end up pasting the same four throws into `AbstractStoreBrokerFactory` rather than deleting them. The methods are also new on the interface (@since 4.2.0, unreleased), so any out-of-tree factory extending `AbstractBrokerFactory` compiles today only because these defaults exist - dropping them turns a clear `UnsupportedOperationException` into an `AbstractMethodError` at `getSchemaManager()` time. I'd rather keep one default here than duplicate it per subclass; this commit only makes the exception say which operation is missing. If you'd still prefer compile-time enforcement, that's a separate change that also has to decide what `AbstractStoreBrokerFactory` does for schema operations.

solomax: Understood, thanks for clarifications :)

solomax (approved):

PR #156, opened 2026-08-20
[OPENJPA-2985] Implement the mandatory JPA 3.2 methods that threw UnsupportedOperationException

rzo1:
Implements the three mandatory JPA 3.2 methods that were still throwing `UnsupportedOperationException("Not yet implemented (JPA 3.2)")`, one commit each with tests: `EntityManager.createQuery(TypedQueryReference)`, `EntityManagerFactory.getNamedQueries(Class)` and `EntityManager.find(EntityGraph, Object, FindOption...)`.

`createQuery(TypedQueryReference)` delegates to `createNamedQuery(name, resultType)` and applies the reference's hints. `getNamedQueries(Class)` needed metadata support first, so `AnnotationPersistenceMetaDataParser` now honours the JPA 3.2 `@NamedQuery.resultClass()` the same way it already honoured `@NamedNativeQuery.resultClass()`; the method then returns a fresh map of `TypedQueryReference` for every named query whose declared result type is assignable to the requested one. `find(EntityGraph, ...)` resolves the root type from the graph and applies it as a **load graph** to the pushed fetch plan (with a cycle guard), which is the first place an `EntityGraph` actually influences loading in OpenJPA — until now `EntityGraphImpl` was a pure data holder.

Two known gaps worth a reviewer's opinion: named queries without a declared result type are excluded from `getNamedQueries` for every requested type including `Object.class`, and because the bundled `orm_3_2.xsd.rsrc` still carries the 3.1-shaped `named-query` type without `result-class`, XML-declared named queries can never be returned — that stale schema looks like a separate follow-up.

### DIFF
--- a/openjpa-persistence/src/main/java/org/apache/openjpa/persistence/EntityManagerImpl.java
+++ b/openjpa-persistence/src/main/java/org/apache/openjpa/persistence/EntityManagerImpl.java
@@ -739,22 +739,23 @@ public class EntityManagerImpl
 	 * @return the lock mode given among the options or null if none was given
 	 */
 	private LockModeType parseFindOptions(Map<String, Object> props, FindOption... options) {
+		if (options == null) {
+			return null;
+		}
 		LockModeType mode = null;
-		if (options != null) {
-			for (FindOption opt: options) {
-				if (opt instanceof LockModeType lmt) {
-					mode = lmt;
-				} else if (opt instanceof CacheRetrieveMode crm) {
-					props.put(JPAProperties.CACHE_RETRIEVE_MODE, crm);
-				} else if (opt instanceof CacheStoreMode csm) {
-					props.put(JPAProperties.CACHE_STORE_MODE, csm);
-				} else if (opt instanceof PessimisticLockScope pls) {
-					props.put(JPAProperties.LOCK_SCOPE, pls);
-				} else if (opt instanceof Timeout timeout) {
-					props.put(JPAProperties.LOCK_TIMEOUT, timeout.milliseconds());
-				}
-				// open to custom options
+		for (FindOption opt: options) {
+			if (opt instanceof LockModeType lmt) {
+				mode = lmt;
+			} else if (opt instanceof CacheRetrieveMode crm) {
+				props.put(JPAProperties.CACHE_RETRIEVE_MODE, crm);
+			} else if (opt instanceof CacheStoreMode csm) {
+				props.put(JPAProperties.CACHE_STORE_MODE, csm);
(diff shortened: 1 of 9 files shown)


solomax:
Build need to be fixed :)

openjpa-persistence/src/main/java/org/apache/openjpa/persistence/EntityManagerImpl.java
```diff
@@ -721,22 +725,38 @@ public <T> T find(Class<T> cls, Object oid,
...
-				mode = lmt;
-			} else if (opt instanceof CacheRetrieveMode crm) {
-				props.put(JPAProperties.CACHE_RETRIEVE_MODE, crm);
-			} else if (opt instanceof CacheStoreMode csm) {
-				props.put(JPAProperties.CACHE_STORE_MODE, csm);
-			} else if (opt instanceof PessimisticLockScope pls) {
-				props.put(JPAProperties.LOCK_SCOPE, pls);
-			} else if (opt instanceof Timeout timeout) {
-				props.put(JPAProperties.LOCK_TIMEOUT, timeout.milliseconds());
+		if (options != null) {
```
solomax: Maybe this one can be flattened a bit with:

```
if (options == null) {
    return mode;
}
```

? :)

openjpa-persistence/src/main/java/org/apache/openjpa/persistence/TypedQueryReferenceImpl.java
```diff
@@ -0,0 +1,99 @@
...
+     * @param hints the query hints, may be null or empty. A defensive, unmodifiable copy is taken.
+     */
+    public TypedQueryReferenceImpl(String name, Class<? extends R> resultType, Map<String, Object> hints) {
+        if (name == null)
+            throw new IllegalArgumentException("name is required");
+        if (resultType == null)
+            throw new IllegalArgumentException("resultType is required");
+        _name = name;
+        _resultType = resultType;
+        _hints = (hints == null || hints.isEmpty())
```
solomax, line 60:
```suggestion
        _hints = (hints == null || hints.isEmpty()) ? Map.of() : Map.copyOf(hints)
```

solomax (approved):
I've added couple suggestions :)

hopefully others can review as well :)

cristof:
LGTM. Does this pass the TCK, @rzo1 , or TCK did not check it?

rzo1:
The TCK passend without that being implemented(that's why it was Left)

cristof (approved):

PR #157, opened 2026-08-20
[OPENJPA-2966] Use a dedicated cast target type for long values

rzo1:
Fixes [OPENJPA-2966](https://issues.apache.org/jira/browse/OPENJPA-2966): casting to `long` used the DDL `decimalTypeName`, which is `DECIMAL(10,0)` on MySQL/MariaDB and silently clamps any value past ten billion. Long casts now go through a new `longCastTypeName`, resolved lazily from `bigintTypeName` (so per-dictionary overrides assigned in constructors or `connectedConfiguration()` are picked up) and set to `SIGNED` in `MySQLDictionary` and `MariaDBDictionary`; no DDL type name is touched, so schema generation is unchanged. The second half of the report is fixed too: `TypecastAsNumber` and `TypecastAsString` now share the size-suffix stripping via `DBDictionary`, which changes exactly one dictionary's output (FoxPro no longer emits the literal `CAST(x AS CHARACTER{0})`).

### DIFF
--- a/openjpa-jdbc/src/main/java/org/apache/openjpa/jdbc/sql/DBDictionary.java
+++ b/openjpa-jdbc/src/main/java/org/apache/openjpa/jdbc/sql/DBDictionary.java
@@ -18,6 +18,8 @@
  */
 package org.apache.openjpa.jdbc.sql;
 
+import static java.util.Locale.ROOT;
+
 import java.io.BufferedReader;
 import java.io.ByteArrayInputStream;
 import java.io.CharArrayReader;
@@ -141,8 +143,6 @@ import org.apache.openjpa.util.Serialization;
 import org.apache.openjpa.util.StoreException;
 import org.apache.openjpa.util.UnsupportedException;
 import org.apache.openjpa.util.UserException;
-
-import static java.util.Locale.ROOT;
 
 
 /**
@@ -518,7 +518,7 @@ public class DBDictionary
      * Dictionaries which change {@link #integerTypeName} do <em>not</em> implicitly change this value.
      * Set it explicitly if the DDL type name is not a valid CAST target.
      */
-    public String integerCastTypeName = integerTypeName;
+    public String integerCastTypeName = null;
 
     /**
      * Type name used as the target of a CAST to a 64 bit integer.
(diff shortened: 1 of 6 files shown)


solomax:
I'll try to carefully check this one tomorrow morning my time :)

openjpa-jdbc/src/main/java/org/apache/openjpa/jdbc/sql/DBDictionary.java
```diff
@@ -511,8 +511,25 @@ public enum DateMillisecondBehaviors { DROP, ROUND, RETAIN }
...
     public boolean supportsUnsizedCharOnCast = true;
 
+    /**
+     * Type name used as the target of a CAST to a 32 bit integer.
+     * <p>
+     * Note: this is a field initializer, so it is evaluated before any subclass constructor runs.
+     * Dictionaries which change {@link #integerTypeName} do <em>not</em> implicitly change this value.
+     * Set it explicitly if the DDL type name is not a valid CAST target.
+     */
     public String integerCastTypeName = integerTypeName;
```
solomax: `integerTypeName` is redefined by some DBDictionaries in constructor (lot's of them actually)
So this one also need to be lazy inited

openjpa-jdbc/src/main/java/org/apache/openjpa/jdbc/sql/DBDictionary.java
```diff
@@ -2223,6 +2240,43 @@ protected int getDateFractionDigits(Column col, String typeName) {
...
+        if (type == int.class || type == Integer.class) {
+            name = integerCastTypeName;
+        } else if (type == long.class || type == Long.class) {
+            name = getLongCastTypeName();
+        } else if (type == float.class || type == Float.class) {
+            name = floatTypeName;
+        } else {
+            name = doubleTypeName;
+        }
+        return insertSize(name, null);
```
solomax: Maybe i worth to cache these values instead of calling `insertSize(...)` all the time it might be performance unwise ....

rzo1: done

solomax (approved):

PR #158, opened 2026-08-21
[OPENJPA-2999] Correct the schema copyright and license statements in the manual

rzo1:
Fixes the `[2024]` template bracket and the orphaned CDDL sentence flagged in https://issues.apache.org/jira/browse/OPENJPA-2999. Checking which of the two statements was stale showed neither was right: the shipped schemas split between Sun/CDDL 1.0 (JPA 1.0 and 2.0) and Oracle `EPL-2.0 OR BSD-3-Clause` (2.1 through 3.2) — EFSL 1.1 covers the spec document, not the schemas. Docs only; the wording is worth a second read since it is legal text.


solomax (approved):
@rmannibucau could you please review this one? :)

cristof (approved):

PR #159, opened 2026-08-21
[OPENJPA-2957] Resolve embeddables through the metadata repository instead of annotation names

rzo1:
Replaces the simple-name annotation scan in `AbstractExpressionBuilder.hasEmbeddableAnnotation` with a repository lookup, as suggested in https://issues.apache.org/jira/browse/OPENJPA-2957. `ClassMetaData.isEmbeddable()` is set by both the annotation and the XML parser, so it also covers embeddables declared only in `orm.xml` and no longer matches any annotation that happens to be called `Embeddable`.

Note that `fmd.getEmbeddedMetaData()` cannot be used directly here: the guard is only reached when that call already returned `null`, so the check has to ask about the declared type instead. This is a robustness fix rather than a behavioural one — I could not construct a case where the old check produced a wrong answer, since the branch is only reachable for an `@EmbeddedId` field whose embedded metadata is absent. `openjpa.persistence.embed.**` (89 tests, including the XML-mapped ones) and the `TestDerivedIdEx2b` / `TestUnenhancedDerivedIdEx2b` regression tests for the original 2.4.1.3 ex2b fix are green.

### DIFF
--- a/openjpa-kernel/src/main/java/org/apache/openjpa/kernel/exps/AbstractExpressionBuilder.java
+++ b/openjpa-kernel/src/main/java/org/apache/openjpa/kernel/exps/AbstractExpressionBuilder.java
@@ -305,8 +305,7 @@ public abstract class AbstractExpressionBuilder {
             meta = fmd.getEmbeddedMetaData();
         else
             meta = fmd.getDeclaredTypeMetaData();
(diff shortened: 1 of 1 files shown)


cristof (approved):

solomax (approved):
LGTM :)

PR #163, opened 2026-08-25
[OPENJPA-2971] Make PostgreSQL StoreCharsAsNumbers default overridable and release-note it

rzo1:
Since #144 `PostgresDictionary.connectedConfiguration()` unconditionally set `storeCharsAsNumbers=false` on PostgreSQL 9+. Because `DBDictionaryFactory` applies the user's `DBDictionary` plugin properties before `connectedConfiguration()`, an explicit `StoreCharsAsNumbers=true` was silently clobbered, breaking schemas created by earlier OpenJPA releases (char fields in `INTEGER` columns).

This adds `setStoreCharsAsNumbers()`/`getStoreCharsAsNumbers()` to `DBDictionary` (`Options` prefers the setter over the public field) to track an explicit user value, makes native CHAR storage the `PostgresDictionary` constructor default, and only falls back to numeric storage on PostgreSQL < 9 when nothing was configured explicitly. The public field is kept for backward compatibility.

Documents the new default and the opt-out (`openjpa.jdbc.DBDictionary=postgres(StoreCharsAsNumbers=true)`) in the reference guide and the JPA 3.2 migration considerations, and adds `TestPostgresDictionary` covering defaults, explicit overrides via plugin properties / setter / field, and `DBDictionaryFactory` end-to-end. The TCK-relevant default (CHAR columns) is unchanged.

Ref: https://github.com/apache/openjpa/pull/144#discussion_r3683003316


openjpa-jdbc/src/main/java/org/apache/openjpa/jdbc/sql/DBDictionary.java
```diff
@@ -315,7 +315,18 @@ public class DBDictionary
...
+     * rather than as <code>CHAR</code> values. Dictionaries may set their own
+     * default and may adjust it in {@link #connectedConfiguration(Connection)}
+     * based on the detected database version; use
+     * {@link #setStoreCharsAsNumbers(boolean)} to configure a value that is
+     * guaranteed to be left untouched (OPENJPA-2971).
+     */
     public boolean storeCharsAsNumbers = true;
+    // true once the user explicitly configured StoreCharsAsNumbers; dictionaries
+    // must not auto-detect over such a value (OPENJPA-2971)
+    private boolean storeCharsAsNumbersExplicit = false;
```
solomax, line 329: Hello @rzo1,
I need some clarifications on this one :)

`storeCharsAsNumbersExplicit`, `getStoreCharsAsNumbers` and `setStoreCharsAsNumbers` seems to be not used in the code

why do we need them?

rzo1: storeCharsAsNumbersExplicit is used, just not directly:
- written in setStoreCharsAsNumbers (DBDictionary.java:6052)
- read by isStoreCharsAsNumbersExplicit() (DBDictionary.java:6062)
- which is called from PostgresDictionary.connectedConfiguration (PostgresDictionary.java:1073): if (maj < 9 && !isStoreCharsAsNumbersExplicit())

Same for the setter: it looks unused because nothing calls it by name, but Options.matchOptionToMember resolves a plugin property to a setter method first and only falls back to the public field, so postgres(StoreCharsAsNumbers=true) goes through it. 

(the getter is unused but there for consistency)

rzo1: It was added to preserve backward compat (as requested by Romain) as backward compat flag for users who have schemas sitting in (upgraded) Postgres databases

solomax: My bad, need more coffee this morning :)
Sorry for the noise :)

solomax (approved):

cristof (approved):

PR #164, opened 2026-08-25
[OPENJPA-2940] Migration considerations for OpenJPA 4.2.0 (Jakarta Persistence 3.2)

rzo1:
Adds a `jpa_4.2` section to `migration_considerations.xml` covering the changes introduced by #144 (Jakarta Persistence 3.2), derived from the merged diff and the review discussion on that PR: platform and dependency requirements, JPQL/query semantics (new reserved identifiers, materialized `getResultList()`, non-cascading bulk `DELETE`, `FlushModeType.AUTO`, rollback-only marking, argument validation), mapping and schema changes (default map key column, attribute overrides, converters, access type, generated DDL, schema generation), database-specific changes (PostgreSQL, MySQL/MariaDB, SQL Server, Oracle, HSQLDB), SPI changes and a compact list of notable new features.

Each item states the previous and the new behaviour and, where one exists, the property that restores the old behaviour. The PostgreSQL `StoreCharsAsNumbers` paragraph describes the behaviour after #163; the corresponding hunk in #163 should then be dropped, since it was placed under the OpenJPA 3.2.0 section.

Draft: the section is long and some per-database / SPI subsections may deserve trimming — reviews welcome on what to keep.


openjpa-project/src/doc/manual/migration_considerations.xml
```diff
@@ -649,5 +649,1081 @@
             </section>
         </section>
     </section>
+    <section id="jpa_4.2">
```
solomax: The only change I would propose: to sort sections in opposite order :) i.e. the newest changes first :)

rzo1: Should we reverse order here or in a separate PR to avpid a huge diff?

solomax: If the idea worth it I would change it in this PR :) 
(or maybe put `jpa_4.2` as first here and move other separately :)))

rzo1: done

solomax (approved):

solomax (approved):
re-approving as requested :))

PR #166, opened 2026-08-29
[OPENJPA-2983] Resolve EXTRACT temporal fields by constant, not by name

rzo1:
`CriteriaBuilderImpl.extract()` derived the `DateTimeExtractField` from `field.toString().toUpperCase(...)`. `Locale.ROOT` on master closed the Turkish-locale hole, but the lookup still depended on the exact wording of the spec constants, and an unrecognised field failed with `Enum.valueOf`'s "No enum constant ..." rather than anything actionable.

The `LocalDateField` / `LocalTimeField` / `LocalDateTimeField` constants are now mapped to their `DateTimeExtractField` equivalents by identity, and anything else is rejected with a meaningful message. Identity is safe: the three classes have a single private constructor, are not `Serializable`, and expose no factory.

No behaviour change for working queries — `LocalDateTimeField.DATE`/`TIME` already threw `IllegalArgumentException` (their `toString()` is lowercase, so `valueOf` never matched) and still do, with a better message.


cristof (approved):

solomax (approved):

solomax (approved):
I do remember I have approved this one already :)

PR #167, opened 2026-08-29
[OPENJPA-2968] Do not swallow failures around IdClass primary key values

rzo1:
The `@IdClass` extraction and reconstruction paths added for the JPA 2.4.1.3 example 2b case caught every exception and continued: on write the primary key columns were filled with nulls, on read the field was left null, so a corrupt identity could be persisted or loaded without a trace.

Both paths now raise a `StoreException` naming the field, the id class and the mapping, with the original failure chained. A null id class value on the write path is legitimate, so it still writes nulls, but logs a warning.

The columns are also no longer paired with the id class fields by `Class.getDeclaredFields()` order, which the JVM does not guarantee. Where every column names its target through `referencedColumnName` the fields are matched by that name; declaration order remains only as the fallback for mappings that provide no name, since that is all such a mapping offers.

`TestDerivedIdEx2c` covers the ordering: its id class declares the fields in the order opposite to the join columns, so a `find()` misses on the previous code and succeeds now.


openjpa-jdbc/src/main/java/org/apache/openjpa/jdbc/meta/strats/EmbedValueHandler.java
```diff
@@ -216,21 +218,27 @@ protected int toDataStoreValue1(OpenJPAStateManager em, ValueMapping vm,
...
-                            for (java.lang.reflect.Field f : df) {
-                                Object fv = f.get(idObj);
-                                if (cols.length == 1) rvals.add(fv);
-                                else ((Object[]) rvals.get(0))[idx++] = fv;
-                            }
-                        } catch (Exception ex) {
-                            for (int c = 0; c < mic.size(); c++) {
-                                if (cols.length == 1) rvals.add(null);
-                                else ((Object[]) rvals.get(0))[idx++] = null;
+                        List<java.lang.reflect.Field> df =
```
solomax: weird enough `java.lang.reflect.Field` is not moved to `import`'s

openjpa-jdbc/src/main/java/org/apache/openjpa/jdbc/meta/strats/EmbedValueHandler.java
```diff
@@ -216,21 +218,27 @@ protected int toDataStoreValue1(OpenJPAStateManager em, ValueMapping vm,
...
+                                    "mapsid-extract-failed", f.getName(),
+                                    idObj.getClass().getName(),
+                                    fms[i].getFullName(false))).setCause(ex);
                             }
+                            if (cols.length == 1) rvals.add(fv);
+                            else ((Object[]) rvals.get(0))[idx++] = fv;
                         }
                     } else {
+                        Log log = fms[i].getRepository().getLog();
+                        if (log.isWarnEnabled())
```
solomax: I would add `{}` here :)

openjpa-jdbc/src/main/java/org/apache/openjpa/jdbc/meta/strats/EmbedValueHandler.java
```diff
@@ -442,20 +462,55 @@ public static void getIdColumns(FieldMapping fmd, List cols) {
...
+        for (java.lang.reflect.Field f : declared) {
+            if (positional.size() >= mic.size()) break;
+            positional.add(f);
+        }
+        return positional;
+    }
+
+    private static java.lang.reflect.Field findField(
+            List<java.lang.reflect.Field> fields, String name) {
+        for (java.lang.reflect.Field f : fields)
```
solomax: I would add `{}` in these `for/if`'s :))

Is the stream API will be slower here? :)
like: `fields.stream().filter(f -> f.getName().equals(name)).findFirst().orElse(null)`

solomax (approved):
Just minor comments :)

openjpa-jdbc/src/main/java/org/apache/openjpa/jdbc/meta/strats/EmbedValueHandler.java
rzo1: Fixed — `Field` and `Modifier` are imported now. The fully qualified names were carried over from the code this replaces.

openjpa-jdbc/src/main/java/org/apache/openjpa/jdbc/meta/strats/EmbedValueHandler.java
rzo1: Added.

openjpa-jdbc/src/main/java/org/apache/openjpa/jdbc/meta/strats/EmbedValueHandler.java
rzo1: Braces added. On the stream: I kept the loops. `findField` does two passes on purpose — exact name first, then case-insensitive — so the stream version would need two `filter(...).findFirst()` chains anyway, and this runs per row on the read path. Happy to switch if you prefer the stream form for readability.

solomax (approved):

PR #168, opened 2026-08-29
[OPENJPA-2967] Make VERSION() work for surrogate versions

rzo1:
Note that the issue description is out of date: `getColumns()` no longer calls `getVersionFieldMapping()` and `initialize()` already validates the class mapping, both since e94a8515b, and both are already covered by tests. What remains is the surrogate case.

`VersionVal.getType()` returned null for a type whose version is mapped by the version strategy rather than by a `@Version` field. `initialize()` admits such a type on purpose, since its version columns exist, so the null escaped into `JPQLExpressionBuilder` (`type.isPrimitive()` for `SELECT VERSION(e)`) and into `Filters.wrap` (for `WHERE VERSION(e) = :v`), both failing with a NullPointerException.

`getType()` now falls back to `Object.class` — the same answer the in-memory `VersionVal` already gives, so the two query paths agree — and `toDataStoreValue()` converts through the version column's java type. Types with a real `@Version` field are unaffected.

`initialize()` also reported an invalid `VERSION()` argument with the `bad-getobjectid` message, which describes an object id failure and whose argument is null for a bare identification variable. It now uses a dedicated message naming the path.


cristof (approved):

openjpa-jdbc/src/main/java/org/apache/openjpa/jdbc/kernel/exps/VersionVal.java
```diff
@@ -78,11 +79,19 @@ public void setMetaData(ClassMetaData meta) {
...
-    	FieldMetaData versionField = _path.getMetaData().getVersionField();
-    	if (versionField != null) {
-    		return versionField.getType();
-    	}
-        return null;
+        // note: getColumns()/initialize() resolve the target type from the
+        // ExpState; getType() has none, so it uses the path's class.
+        ClassMetaData meta = _path.getMetaData();
+        FieldMetaData versionField = (meta == null) ? null : meta.getVersionField();
+        if (versionField != null)
```
solomax: Could you please add `{}`? :)

openjpa-jdbc/src/main/java/org/apache/openjpa/jdbc/kernel/exps/VersionVal.java
```diff
@@ -98,23 +107,43 @@ public ExpState initialize(Select sel, ExpContext ctx, int flags) {
...
         ClassMapping cls = _path.getClassMapping(state);
         if (cls == null || cls.getEmbeddingMapping() != null)
-            throw new UserException(_loc.get("bad-getobjectid", _path.getFieldMapping(state)));
+            throw new UserException(_loc.get("bad-version-path", pathDescription()));
 
         // types that are not versioned have no version columns to select,
         // group, order or compare by; fail with a meaningful message rather
         // than a NullPointerException further down the line
-        if (cls.getVersion() == null || cls.getVersion().getColumns().length == 0)
+        if (cls.getVersion().getColumns().length == 0)
```
solomax: `cls.getVersion()` can't be `null` here?

openjpa-jdbc/src/main/java/org/apache/openjpa/jdbc/kernel/exps/VersionVal.java
solomax: In all changes :)

solomax (changes requested):

openjpa-jdbc/src/main/java/org/apache/openjpa/jdbc/kernel/exps/VersionVal.java
rzo1: Done, every block this PR touches.

openjpa-jdbc/src/main/java/org/apache/openjpa/jdbc/kernel/exps/VersionVal.java
rzo1: Correct, it cannot. `ClassMapping._version` is `private final`, assigned unconditionally in both constructors (`ClassMapping.java:102` and `:112`, via `MappingRepository.newVersion()` which always returns `new Version(cls)`), there is no setter, and there's no subclass of `ClassMapping`/`MappingRepository` or override of `newVersion()`/`getVersion()` in the tree. An unresolved mapping still has one — `_version.resolve(...)` runs later and never clears it — and the embeddable case is rejected one line earlier by the `getEmbeddingMapping() != null` check. What an unversioned type gives you is a `Version` with no columns, and `getColumns()` is null-safe in itself (`_cols` starts as `EMPTY_COLUMNS`, `setColumns()` coerces null back), so the remaining `getColumns().length == 0` is the only check that means anything.

solomax (approved):

PR #169, opened 2026-08-29
[OPENJPA-2986] Let setTimeout(null) clear a query timeout

rzo1:
`Query.setTimeout(null)` passed the null straight to `setHint()`, which drove it through `FetchConfigurationImpl.setHint()` into a reflective call of `setQueryTimeout(int)` and failed with a `bad-hint-value` `IllegalArgumentException`. So a timeout set through this API could never be cleared — on queries it did not silently ignore the null, it threw — and `getTimeout()` kept reporting the stale value.

A null timeout now restores the timeout the query inherits from its entity manager, which is what the query would have used had `setTimeout()` never been called. Deliberately not `FetchPlan.DEFAULT`, which resolves to the `OpenJPAConfiguration` value and would discard an entity-manager-scoped timeout the caller never set here.

`getTimeout()` reports the effective fetch plan value rather than the recorded hint, so it cannot disagree with what the query will actually use, and `StoredProcedureQueryImpl` delegates both methods to its query so the two cannot drift apart.

`EntityManagerImpl` carries a `setTimeout`/`getTimeout` pair too, but those implement `EntityTransaction` — a transaction timeout in seconds — and are left alone here.


cristof (approved):

openjpa-persistence/src/main/java/org/apache/openjpa/persistence/QueryImpl.java
```diff
@@ -780,20 +780,28 @@ public TypedQuery<X> setCacheStoreMode(CacheStoreMode cacheStoreMode) {
...
-		if (val instanceof Number) {
-			return ((Number) val).intValue();
-		}
-		return null;
+		int timeout = getFetchPlan().getQueryTimeout();
+		return timeout > 0 ? timeout : null;
 	}
 
+	/**
+	 * Sets the query timeout in milliseconds. A null timeout clears a timeout
```
solomax: Is this comment accurate? it states: `A null timeout clears a timeout set through this method`

openjpa-persistence/src/main/java/org/apache/openjpa/persistence/QueryImpl.java
```diff
@@ -780,20 +780,28 @@ public TypedQuery<X> setCacheStoreMode(CacheStoreMode cacheStoreMode) {
...
 	public Integer getTimeout() {
-		Object val = getHints().get(JPAProperties.QUERY_TIMEOUT);
-		if (val instanceof Integer) {
-			return (Integer) val;
-		}
-		if (val instanceof Number) {
-			return ((Number) val).intValue();
-		}
-		return null;
+		int timeout = getFetchPlan().getQueryTimeout();
```
solomax: `set` method put timeout to the `hind` while `get` takes it from `fetchPlan`
Is this OK?

rzo1: Good catch, it was not OK. The write path is fine — the value reliably reaches the plan, and the JPA key is the highest-ranked one so the precedence check can't swallow it — but deriving null from `timeout > 0` was wrong both ways: an explicit `setTimeout(0)` read back as null, and a query that only *inherited* a timeout reported one that was never set on it. `getTimeout()` now returns null when nothing is set on the query and the effective value otherwise, so both sides answer from the same source.

Chasing this also turned up a side effect of the old `setTimeout(null)`: it re-recorded the inherited value as an explicit `jakarta.persistence.query.timeout` hint, which made the highest-ranked key "set" and silently turned a later `setHint("openjpa.FetchPlan.QueryTimeout", n)` into a no-op. It now drops the hint instead.

openjpa-persistence/src/main/java/org/apache/openjpa/persistence/QueryImpl.java
rzo1: It wasn't, on two counts, and it's rewritten. `setTimeout(null)` clears the timeout whatever set it — `setHint("jakarta.persistence.query.timeout", n)` or `getFetchPlan().setQueryTimeout(n)` just as much as `setTimeout(n)` — and it restores the value the entity manager carries *at that point*, not the one the query was created with: the query's fetch config is cloned from the broker at creation, while `inheritedQueryTimeout()` reads the EM live.

solomax (approved):

PR #170, opened 2026-08-29
[OPENJPA-2958] Add ImplHelper.registerUnenhancedInstance()

rzo1:
`BrokerImpl` reached into the public mutable `ImplHelper._unenhancedInstanceMap` to register the persistence-capable instance of an unenhanced one. This adds a method for it and routes all three write sites through it, so the registry is no longer written to from outside the class that owns it. Behaviour is unchanged — the method is the same overwriting put.

The field itself is now package private. Nothing outside `ImplHelper` reads or writes it any more, and 4.2.0 already removes public API elsewhere, so keeping it public for compatibility was not worth it. Package private rather than private so the unit test, which lives in the same package, can still read a registration back: there is no public read path, since `toPersistenceCapable()` rejects a non-manageable instance before it reaches the map.


cristof (approved):

openjpa-kernel/src/main/java/org/apache/openjpa/util/ImplHelper.java
```diff
@@ -61,7 +61,16 @@ public class ImplHelper {
...
+     * Map of all new unenhanced instances active in this classloader, keyed
+     * by object identity. Keys are held weakly, values strongly.
+     *
+     * @deprecated use
+     * {@link #registerUnenhancedInstance(Object, PersistenceCapable)} to add
+     * mappings and {@link #toPersistenceCapable(Object, Object)} to resolve
+     * them.
+     */
+    @Deprecated
     public static final Map _unenhancedInstanceMap =
```
solomax: We are already add some incompatibilities maybe it worth to make it `private`?

solomax (approved):

openjpa-kernel/src/main/java/org/apache/openjpa/util/ImplHelper.java
rzo1: Agreed, and my reasoning in the description was wrong — 4.2.0 already removes public API (`J2DoPrivHelper` lost ~52 public static methods, `DBDictionary.SerializedData` became a record and dropped a public field), so there was nothing left to protect. Nothing outside `ImplHelper` touches the field any more.

I made it package private rather than private: once it's private there's no read path at all, since `toPersistenceCapable()` rejects a non-manageable instance before it reaches the map, so the unit test in the same package could no longer verify a registration. Description updated.

solomax (approved):

PR #184, opened 2026-09-14
[OPENJPA-2970] Let an explicit null named parameter win over positional lookup

rzo1:
Stored procedure parameters are now looked up with `containsKey`, so a named parameter bound to null is no longer replaced by the value registered under its position. The parameter array is now indexed by position, so an IN parameter that follows an OUT parameter gets bound instead of being silently skipped. New tests in `TestStoredProcedureQueryFixes` cover both cases.


openjpa-jdbc/src/main/java/org/apache/openjpa/jdbc/kernel/StoredProcedureQuery.java
```diff
@@ -186,15 +186,17 @@ public ResultObjectProvider executeQuery(StoreQuery q, Object[] params, Range ra
                 stmnt = conn.prepareCall(_proc.getCallSQL());
 
                 final StoredProcedureQuery spq = (StoredProcedureQuery) q;
+                // no parameter values at all: leave the (un)binding to the driver
+                final boolean bind = params != null && params.length > 0;
                 for (Column c : spq.getProcedure().getInColumns()) {
-                    if (params != null && c.getIndex() < params.length) {
+                    if (bind) {
```
solomax, line 192: `c.getIndex() < params.length` was dropped in 2 code blocks
Are you sure this check is not necessary anymore?

PR #186, opened 2026-09-14
[OPENJPA-2977] Stop mutating PersistenceConfiguration and honour mapping files

rzo1:
`PersistenceUnitInfoImpl.convert()` no longer writes into the caller's `PersistenceConfiguration`; the derived OpenJPA properties go into a copy, and `mappingFiles()` are now wired into the persistence unit. This also fixes merging a user supplied `openjpa.MetaDataFactory`, which broke when creating a second EMF from the same configuration, and drops the unused `PersistenceProductDerivation.load(PersistenceConfiguration, Map)`.


openjpa-persistence/src/main/java/org/apache/openjpa/persistence/PersistenceProductDerivation.java
solomax, line 40:
```suggestion
import jakarta.persistence.PersistenceUnitTransactionType;
```

openjpa-persistence/src/main/java/org/apache/openjpa/persistence/PersistenceUnitInfoImpl.java
solomax, line 42:
```suggestion
import jakarta.persistence.PersistenceUnitTransactionType;
```

openjpa-persistence/src/main/java/org/apache/openjpa/persistence/PersistenceUnitInfoImpl.java
```diff
@@ -650,17 +655,40 @@ public static PersistenceUnitInfoImpl convert(PersistenceConfiguration config) {
...
 				PersistenceUnitTransactionType.JTA : PersistenceUnitTransactionType.RESOURCE_LOCAL);
 		pinfo.setValidationMode(config.validationMode());
-		List<Class<?>> managedClasses = config.managedClasses();
-		if (managedClasses != null && !managedClasses.isEmpty()) {
-			String managedClassesList = managedClasses.stream().map(Class::getName).collect(Collectors.joining(";"));
-			String old = config.properties().containsKey("openjpa.MetaDataFactory")
-					? "," + config.properties().get("openjpa.MetaDataFactory").toString()
-					: "";
-			config.property("openjpa.MetaDataFactory", "jpa(Types=" + managedClassesList + old + ")");
+		for (Class<?> managedClass : config.managedClasses()) {
```
solomax, line 658: null check is dropped here, is it OK?

openjpa-persistence/src/main/java/org/apache/openjpa/persistence/PersistenceUnitInfoImpl.java
```diff
@@ -650,17 +655,40 @@ public static PersistenceUnitInfoImpl convert(PersistenceConfiguration config) {
...
+		}
+		if (!metaFactoryProps.isEmpty()) {
+			// the user properties take precedence over the ones derived from the unit info,
+			// so merge the locations into any user provided metadata factory
+			String key = ProductDerivations.getConfigurationKey("MetaDataFactory", props);
+			Object old = props.get(key);
+			props.put(key, Configurations.combinePlugins(old == null ? null : old.toString(),
+					Configurations.serializeProperties(metaFactoryProps)));
+		}
+		props.put("openjpa.noPersistenceXMLResource", true);
```
solomax: this one is dropped in next PR ....

openjpa-persistence/src/main/java/org/apache/openjpa/persistence/PersistenceUnitInfoImpl.java
solomax: I mean `config.managedClasses()` was previously checked for `null`

rzo1: `PersistenceConfiguration` sets `managedClasses` (and `mappingFileNames`) up as `private final List<...> = new ArrayList<>()` and `managedClasses()`/`mappingFiles()` return those lists directly, so they can't be null. Only a subclass overriding the accessor to return `null` could break that, and that would violate the API. The old code didn't null-check `config.properties()` either.

solomax (approved):

cristof (approved):

PR #187, opened 2026-09-14
[OPENJPA-2977] Drop the unrecognized openjpa.noPersistenceXMLResource property

rzo1:
Drops `openjpa.noPersistenceXMLResource`, which OpenJPA does not recognize and which only logged a "not recognized and will be ignored" warning on every EMF created from a `PersistenceConfiguration`.


openjpa-persistence/src/main/java/org/apache/openjpa/persistence/PersistenceUnitInfoImpl.java
solomax, line 42:
```suggestion
import jakarta.persistence.PersistenceUnitTransactionType;
```

cristof (approved):

PR #189, opened 2026-09-14
[OPENJPA-2961] Include subclasses of the treated class in the TREAT discriminator filter

rzo1:
The TREAT discriminator filter only matched the exact treated class (`disc = value`), so `TREAT(x AS Middle)` dropped instances of Middle's subclasses. It now reuses the discriminator's class conditions, like `InstanceofExpression`, which emit `IN (...)` over the treated class and its subclasses. Note that for `LEFT JOIN TREAT(...)` rows without a related entity are now kept (`disc IS NULL OR ...`); `TestTreatSubclassDiscriminator` covers a three-level single-table hierarchy.


openjpa-persistence-jdbc/src/test/java/org/apache/openjpa/persistence/jpql/treatjoinon/TestTreatSubclassDiscriminator.java
```diff
@@ -0,0 +1,93 @@
...
+            li.setOrder(order);
+            em.persist(li);
+        }
+
+        em.getTransaction().commit();
+        em.close();
+    }
+
+    public void testTreatJoinIncludesSubclasses() {
+        EntityManager em = emf.createEntityManager();
```
solomax: I would use `try (EntityManager em = emf.createEntityManager()) {`

for new code :)

solomax (approved):

cristof:
@rzo1 , could you add, in your tests, a revision number for both software and game, and a genre for game, and add the following tests? It seems to fail when using WHERE TREAT:

```
   public void testTreatInWhereExclusiveProperty() {
        try (EntityManager em = emf.createEntityManager()){
            List<String> results = em.createQuery(
                "SELECT p.name FROM TProduct p WHERE TREAT(p AS TGameProduct).genre = 'FPS'",
                String.class).getResultList();
            assertEquals(List.of("Game"), results);
        } 
    }

    public void testTreatInWhereCommonProperty() {
        try (EntityManager em = emf.createEntityManager()){
            List<String> results = em.createQuery(
                "SELECT p.name FROM TProduct p WHERE TREAT(p AS TGameProduct).revisionNumber = 2.0",
                String.class).getResultList();
            assertEquals(List.of("Game"), results);
        } 
    }
```

rzo1:
Thanks, good catch. The exclusive-property case only passed because `genre` is null on the other rows; the common-property one returned `[Software, Game]`. A TREAT path in WHERE didn't restrict the type at all. A predicate over `TREAT(x AS Type)` is now false when `x` isn't a `Type` or one of its subclasses (spec 4.4.9). The restriction is applied to that predicate only, so other OR branches still match. I added both of your tests, plus one for an OR condition and one for TREAT on a join variable.

cristof (approved):

cristof:
@solomax, as I've added some code, could you please review and merge?

solomax (approved):

PR #190, opened 2026-09-15
[OPENJPA-3005] Use jakarta.persistence.PersistenceUnitTransactionType internally

rzo1:
https://issues.apache.org/jira/browse/OPENJPA-3005

Follow-up to the review of #186: `jakarta.persistence.spi.PersistenceUnitTransactionType` is deprecated for removal in Jakarta Persistence 3.2, but the imports cannot simply be swapped because `PersistenceUnitInfo#getTransactionType()` still returns the SPI enum until the next major version.

- `PersistenceUnitInfoImpl` keeps the transaction type as `jakarta.persistence.PersistenceUnitTransactionType` and only converts to the SPI enum in `getTransactionType()`. Once the spec changes the return type, only that getter needs to change.
- A deprecated `setTransactionType` overload taking the SPI enum is kept for compatibility.
- The `jakarta.persistence.transactionType` property accepts Strings and both enums. Passing a `jakarta.persistence.PersistenceUnitTransactionType` value used to be silently turned into a `null` transaction type.
- `convert(PersistenceConfiguration)` no longer needs the manual JTA/RESOURCE_LOCAL mapping.
- `PersistenceProductDerivation` uses the new enum when parsing `transaction-type` from `persistence.xml`.

`TestPersistenceUnitInfoTransactionType` covers the default, `PersistenceConfiguration`, String / enum / SPI enum properties and the deprecated setter. `testFromUserPropertiesAsEnum` fails without this change.

Note: this touches the import block of `PersistenceProductDerivation` next to a line removed by #186, so whichever is merged second needs a trivial rebase.


solomax (approved):
sorry for initiating this
I was sure it should be easier :(

cristof (approved):

